PR #12, opened 2023-02-08
Transform-to-and-from-gotext

DarjaLeonova:
Added model and convert directories. Added from and to go conversion and tests for it. Updated Earthfile.

### DIFF
--- a/pkg/convert/go.go
+++ b/pkg/convert/go.go
@@ -9,9 +9,9 @@ import (
 )
 
 func ToGo(m model.Messages) ([]byte, error) {
-	pipelineMsg := messagesToPipeline(m)
+	pipelineMsgs := messagesToPipeline(m)
 
-	msg, err := json.Marshal(pipelineMsg)
+	msg, err := json.Marshal(pipelineMsgs)
 	if err != nil {
 		return nil, fmt.Errorf("encode pipeline.Messages to JSON: %w", err)
 	}
@@ -20,16 +20,13 @@ func ToGo(m model.Messages) ([]byte, error) {
 }
 
 func FromGo(b []byte) (model.Messages, error) {
-	var pipelineMsg pipeline.Messages
+	var pipelineMsgs pipeline.Messages
 
-	err := json.Unmarshal(b, &pipelineMsg)
-	if err != nil {
-		return model.Messages{}, fmt.Errorf("decode JSON to pipeline.Messages: %w", err)
+	if err := json.Unmarshal(b, &pipelineMsgs); err != nil {
+		return model.Messages{}, fmt.Errorf("failed to decode JSON to pipeline.Messages: %w", err)
 	}
 
-	msg := messagesFromPipeline(pipelineMsg)
-
-	return msg, nil
+	return messagesFromPipeline(pipelineMsgs), nil
 }
 
 func messagesToPipeline(m model.Messages) pipeline.Messages {
@@ -40,10 +37,10 @@ func messagesToPipeline(m model.Messages) pipeline.Messages {
 
 	for _, value := range m.Messages {
 		pipelineMsg.Messages = append(pipelineMsg.Messages, pipeline.Message{
-			ID:      pipeline.IDList{value.ID},
-			Message: pipeline.Text{Msg: value.Message},
-			Meaning: value.Description,
-			Fuzzy:   value.Fuzzy,
+			ID:          pipeline.IDList{value.ID},
+			Translation: pipeline.Text{Msg: value.Message},
+			Meaning:     value.Description,
+			Fuzzy:       value.Fuzzy,
 		})
 	}
 
@@ -61,7 +58,7 @@ func messagesFromPipeline(m pipeline.Messages) model.Messages {
 			ID:          value.ID[0],
 			Fuzzy:       value.Fuzzy,
 			Description: value.Meaning,
-			Message:     value.Translation.Msg,
+			Message:     value.Message.Msg,
 		})
 	}
 
--- a/pkg/convert/go.go
+++ b/pkg/convert/go.go
@@ -1,0 +1,69 @@
+package convert
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"go.expect.digital/translate/pkg/model"
+	"golang.org/x/text/message/pipeline"
+)
+
+func ToGo(m model.Messages) ([]byte, error) {
+	pipelineMsg := messagesToPipeline(m)
+
+	msg, err := json.Marshal(pipelineMsg)
+	if err != nil {
+		return nil, fmt.Errorf("encode pipeline.Messages to JSON: %w", err)
+	}
+
+	return msg, nil
+}
+
+func FromGo(b []byte) (model.Messages, error) {
+	var pipelineMsg pipeline.Messages
+
+	err := json.Unmarshal(b, &pipelineMsg)
+	if err != nil {
+		return model.Messages{}, fmt.Errorf("decode JSON to pipeline.Messages: %w", err)
+	}
+
+	msg := messagesFromPipeline(pipelineMsg)
+
+	return msg, nil
+}
+
+func messagesToPipeline(m model.Messages) pipeline.Messages {
+	pipelineMsg := pipeline.Messages{
+		Language: m.Language,
+		Messages: make([]pipeline.Message, 0, len(m.Messages)),
+	}
+
+	for _, value := range m.Messages {
+		pipelineMsg.Messages = append(pipelineMsg.Messages, pipeline.Message{
+			ID:      pipeline.IDList{value.ID},
+			Message: pipeline.Text{Msg: value.Message},
+			Meaning: value.Description,
+			Fuzzy:   value.Fuzzy,
+		})
+	}
+
+	return pipelineMsg
+}
+
+func messagesFromPipeline(m pipeline.Messages) model.Messages {
+	msg := model.Messages{
+		Language: m.Language,
+		Messages: make([]model.Message, 0, len(m.Messages)),
+	}
+
+	for _, value := range m.Messages {
+		msg.Messages = append(msg.Messages, model.Message{
+			ID:          value.ID[0],
+			Fuzzy:       value.Fuzzy,
+			Description: value.Meaning,
+			Message:     value.Translation.Msg,
+		})
+	}
+
+	return msg
+}
--- a/pkg/convert/go_test.go
+++ b/pkg/convert/go_test.go
@@ -54,7 +54,9 @@ func TestToGo(t *testing.T) {
 
 	buffer := new(bytes.Buffer)
 
-	assert.NoError(t, json.Compact(buffer, expected))
+	if !assert.NoError(t, json.Compact(buffer, expected)) {
+		return
+	}
 
 	result, err := ToGo(modelMsg)
 
@@ -87,7 +89,9 @@ func TestFromGo(t *testing.T) {
 
 	buffer := new(bytes.Buffer)
 
-	assert.NoError(t, json.Compact(buffer, b))
+	if !assert.NoError(t, json.Compact(buffer, b)) {
+		return
+	}
 
 	result, err := FromGo(buffer.Bytes())
 
--- a/pkg/convert/go_test.go
+++ b/pkg/convert/go_test.go
@@ -1,0 +1,74 @@
+package convert
+
+import (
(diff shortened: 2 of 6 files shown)


pkg/convert/go.go
```diff
@@ -0,0 +1,69 @@
+package convert
```
klavinsjanis, line 1: strange filename go.go

DarjaLeonova: We were using the same file name in POC.

pkg/convert/go_test.go
```diff
@@ -0,0 +1,74 @@
...
+		assert.Error(t, err)
+	}
+
+	assert.Equal(t, expectedByteResult, actualByteResult)
+}
+
+func Test_FromGo_ReturnsExpectedModel(t *testing.T) {
+	t.Parallel()
+
+	actualMessages, err := FromGo(expectedByteResult)
```
klavinsjanis, line 68: messages, err := ...

pkg/convert/go.go
```diff
@@ -0,0 +1,69 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"go.expect.digital/translate/pkg/model"
+	"golang.org/x/text/message/pipeline"
+)
+
+func ToGo(m model.Messages) ([]byte, error) {
```
klavinsjanis, line 11: Maybe there is a better way to name these functions?
 `ToGo()`  sounds a bit odd.

DarjaLeonova: We were using the same functions names in POC.

VladislavsPerkanuks: I thought about that, maybe we can use convention `MessagesToGo` and `GoToMessages` and same for other languages `MessagesToNgx` and `NgxToMessages`, as I saw in other packages they are using some similar naming conventions for  convertions

pkg/convert/go.go
```diff
@@ -0,0 +1,69 @@
...
+	if err != nil {
+		return model.Messages{}, fmt.Errorf("decode JSON to pipeline.Messages: %w", err)
+	}
+
+	msg := messagesFromPipeline(pipelineMsg)
+
+	return msg, nil
+}
+
+func messagesToPipeline(m model.Messages) pipeline.Messages {
```
VladislavsPerkanuks, line 35: do we really need, two additonal functions for that ? e.g. messagesToPipeline and messagesFromPipeline, imo it decreases readability as namings for all functions are too similar to each other and that way makes it more confusing

pkg/convert/go.go
janishorsts: We had the following transformations for different use cases:

- transform from/to GRPC messages (package transform)
- transform from/to files (package convert)

I would prefer to have a separate package for each one. Package name suggestions are welcome.

pkg/convert/go.go
```diff
@@ -0,0 +1,69 @@
...
+	var pipelineMsg pipeline.Messages
+
+	err := json.Unmarshal(b, &pipelineMsg)
+	if err != nil {
+		return model.Messages{}, fmt.Errorf("decode JSON to pipeline.Messages: %w", err)
+	}
+
+	msg := messagesFromPipeline(pipelineMsg)
+
+	return msg, nil
```
klavinsjanis, line 32: return messagesFromPipeline(pipelineMsg), nil

pkg/convert/go_test.go
```diff
@@ -0,0 +1,74 @@
...
+	expectedByteResult, _ = json.Marshal(messagesToPipeline(inputMsg)) //nolint:errchkjson
+)
+
+func Test_ToGo_ReturnsExpectedByteSlice(t *testing.T) {
+	t.Parallel()
+
+	actualByteResult, err := ToGo(inputMsg)
+	if err != nil {
+		assert.Error(t, err)
+	}
```
VladislavsPerkanuks, line 60: redundant if, could be changed to assert.NoError(t,err)

pkg/convert/go.go
```diff
@@ -0,0 +1,69 @@
...
+		Language: m.Language,
+		Messages: make([]model.Message, 0, len(m.Messages)),
+	}
+
+	for _, value := range m.Messages {
+		msg.Messages = append(msg.Messages, model.Message{
+			ID:          value.ID[0],
+			Fuzzy:       value.Fuzzy,
+			Description: value.Meaning,
+			Message:     value.Translation.Msg,
```
janishorsts, line 64: It's the other way around:
* When we import text for translation, we read it from **Message**
* When we export the translated text, we write it to **Translation**

pkg/convert/go_test.go
```diff
@@ -0,0 +1,74 @@
...
+
+	assert.Equal(t, expectedByteResult, actualByteResult)
```
janishorsts, line 62: The test is fragile - the expected value is created by the implementation which may or may not be correct. Avoid it! Test against explicit values.

pkg/convert/go.go
janishorsts: The narrative of the package is very narrow - only to serialize messages from/to different file formats. We won't transform other data structures in this package.

I believe we'll come up with a very good package name in the future which reflects the intent and narrative.

The package name is the culprit. I'm out of ideas at the moment.

Also, please note, we are not gonna make it right from the first go. We are open to changes.

pkg/convert/go.go
DarjaLeonova: I'm not sure that it is good idea because if we will combine these two functions then first of all the function length will be too long which isn't good practice according to clean code book ;D. The second we will mix two functionalities which is also not great.

pkg/convert/go.go
```diff
@@ -0,0 +1,69 @@
...
+	pipelineMsg := pipeline.Messages{
+		Language: m.Language,
+		Messages: make([]pipeline.Message, 0, len(m.Messages)),
+	}
+
+	for _, value := range m.Messages {
+		pipelineMsg.Messages = append(pipelineMsg.Messages, pipeline.Message{
+			ID:      pipeline.IDList{value.ID},
+			Message: pipeline.Text{Msg: value.Message},
+			Meaning: value.Description,
```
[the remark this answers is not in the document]
DarjaLeonova: I checked the source code: https://cs.opensource.google/go/x/text/+/master:message/pipeline/message.go;l=40?q=Meaning&ss=go 
Also, I have checked the links that you sent, but I wasn't able to find the idea about the "Comment" and "Meaning" fields even in library technical documentation these fields are skipped without any comments.
I think that "Meaning" is the meaning of our message and there are 2 comments (Comment and TranslatorComment both of type string). I suppose that "Comment" and "TranslatorComment" are for adding some sort of information like status or something else...

pkg/convert/go_test.go
```diff
@@ -0,0 +1,89 @@
...
+	tests := []struct {
+		name    string
+		m       model.Messages
+		want    []byte
+		wantErr bool
+	}{
+		{
+			name:    "When input is correct, then return expected result",
+			m:       modelMsg,
+			want:    []byte(`{"language":"en","messages":[{"id":"1","meaning":"description1","message":"","translation":"message1","fuzzy":true},{"id":"2","meaning":"description2","message":"","translation":"message2"}]}`), //nolint:lll
```
DarjaLeonova, line 42: Maybe we can export this long []byte to the model object pipeline.Messages and use json.Marshal? Because I think it would be more readable and shorter.

pkg/convert/go.go
DarjaLeonova: Can you check if it is correct after I have pushed my fixes? ;D

pkg/convert/go_test.go
janishorsts: It's up to you. Personally, I would use a byte slice.

janishorsts (changes requested):

pkg/convert/go.go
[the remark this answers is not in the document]
VladislavsPerkanuks: ChatGPT for the rescue
image
I did not check elsewhere, so take it with grain of salt, but it seems appropriate

pkg/convert/go_test.go
VladislavsPerkanuks: You could do
```go
[]byte(`{
	"language":"lv-lv",
	"messages":[
		 {
				"id":"1",
				"meaning":"When you great someone",
				"message":"hello",
				"translation":"čau",
				"fuzzy":false
		 }
	]
}`,
)
```

but with correct identation

pkg/convert/go_test.go
```diff
@@ -0,0 +1,62 @@
...
+		want []byte
+	}{
+		m:    modelMsg,
+		want: []byte(`{"language":"en","messages":[{"id":"1","meaning":"description1","message":"","translation":"message1","fuzzy":true},{"id":"2","meaning":"description2","message":"","translation":"message2"}]}`), //nolint:lll
+	}
+
+	result, err := ToGo(test.m)
+
+	assert.NoError(t, err)
+	assert.Equal(t, test.want, result)
```
janishorsts, line 44: This improves readability by removing all the clutter from the test tables.

```suggestion
	expected := []byte(`{"language":"en","messages":[{"id":"1","meaning":"description1","message":"","translation":"message1","fuzzy":true},{"id":"2","meaning":"description2","message":"","translation":"message2"}]}`) //nolint:lll
	result, err := ToGo(modelMsg)

	assert.NoError(t, err)
	assert.Equal(t, , expected, result)
```

pkg/convert/go_test.go
```diff
@@ -0,0 +1,67 @@
...
+		},
+	},
+}
+
+func TestToGo(t *testing.T) {
+	t.Parallel()
+
+	expected := []byte(`{"language":"en","messages":
+[{"id":"1","meaning":"description1","message":"","translation":"message1","fuzzy":true},
+{"id":"2","meaning":"description2","message":"","translation":"message2"}]}`)
```
VladislavsPerkanuks, line 38: If you want to use `Compact` you could also write fully intended json which is more readable
```suggestion
	expected := []byte(`
	{
		"language":"en",
		"messages":[
			{
				"id":"1",
				"meaning":"description1",
				"message":"",
				"translation":"message1",
				"fuzzy":true
			},
			{
				"id":"2",
				"meaning":"description2",
				"message":"",
				"translation":"message2"
			}
		]
	}`)
```

pkg/convert/go_test.go
```diff
@@ -0,0 +1,67 @@
...
+
+	buffer := new(bytes.Buffer)
+	if err := json.Compact(buffer, expected); err != nil {
+		fmt.Println(err)
```
VladislavsPerkanuks, line 42: You most probably want to assert error, because otherwise when the test will fail, there will be no straight answer on why it failed, due `ToGo` or due  `json.Compact`

klavinsjanis: Same goes for TestFromGo()

DarjaLeonova: done ;)

pkg/convert/go_test.go
```diff
@@ -0,0 +1,98 @@
...
+				"meaning":"description2",
+				"message":"",
+				"translation":"message2"
+			}
+		]
+	}`)
+
+	buffer := new(bytes.Buffer)
+	if err := json.Compact(buffer, expected); err != nil {
+		assert.Error(t, err)
```
klavinsjanis, line 57: assert.NoError(t, json.Compact(buffer, expected))

klavinsjanis (approved):

pkg/convert/go_test.go
```diff
@@ -0,0 +1,96 @@
...
+				"meaning":"description2",
+				"message":"",
+				"translation":"message2"
+			}
+		]
+	}`)
+
+	buffer := new(bytes.Buffer)
+
+	assert.NoError(t, json.Compact(buffer, expected))
```
VladislavsPerkanuks, line 57:
```suggestion
	if !assert.NoError(t, json.Compact(buffer, expected)){
		return
	}
```
That way if there is an error in `json.compact` test will not continue (it is pointless to continue as everything else will likely fail, the final assertion for sure), so in a result, if this is the reason why the test is failing, it will output only 1 failed assertion instead of x failed assertions. In that way, it pinpoints the exact place where everything has gone wrong.

DarjaLeonova: Thank u!

VladislavsPerkanuks (approved):

janishorsts (approved):

PR #13, opened 2023-02-08
File-upload2

VladislavsPerkanuks:


Earthfile
```diff
@@ -20,11 +20,26 @@ proto:
...
+	\t\t}\n\
+	\n\
+	\tdata, err := io.ReadAll(file)\n\
+	\t\tif err != nil {\n\
+		\t\t\treturn nil, metadata, status.Errorf(codes.InvalidArgument, "%v", err)\n\
+	\t\t}\n\
+	\n\
+	\tprotoReq.Data = data\n' gen/proto/go/translate/v1/translate.pb.gw.go
+
+  RUN rm gen/proto/go/translate/v1/translate.pb.gw.go.bu
```
VladislavsPerkanuks, line 37: That is the best solution I could think of. So bassicaly we search and add code, in generated code by grpc-gateway.  That saves us writing custom handlers, and curl is executed using form data e.g.:
```curl
curl --location --request PUT 'http://localhost:8080/v1/files/' \
--form 'file=@"/Users/vladislavs/Projects/translate/example.translate.json"'
```

pkg/translate/translate.go
```diff
@@ -0,0 +1,44 @@
...
+	req *pb.UploadTranslationFileRequest,
+) (*pb.UploadTranslationFileResponse, error) {
+	var (
+		reqLanguage = req.GetLanguage()
+		reqData     = req.GetData()
+		reqSchema   = req.GetSchema()
+	)
+
+	if len(reqLanguage) == 0 {
+		return nil, status.Errorf(codes.InvalidArgument, "'language' is missing")
```
janishorsts, line 27:
```suggestion
		return nil, status.Errorf(codes.InvalidArgument, "'language' is required")
```

janishorsts (changes requested):
Integration tests are required. If you'd like help with them, please let me know.

Earthfile
janishorsts: We need a comment to describe why you do it.

VladislavsPerkanuks: Basically, I have identified 4 options:

1. Develop a middleware that reads the file from the form-data and embeds it in the request body (not feasible due to various limitations, primarily related to the grpc-gateway implementation).
2. Divide the upload request into two endpoints, one public and one internal. The idea is that the client sends a request to the external endpoint with the file in the form-data, then that endpoint reads the file and makes a request to the internal endpoint. (This would require custom handlers and could cause issues with traces in the future, and would result in a total of three calls instead of two).
3. Similar to option 2, but with only one endpoint for the REST API. This would read the file from the form-data and then send the request in gRPC (which would require developing a custom handler and a client for gRPC for just one use case, which seemed overly cumbersome).
4. Modifying the grpc-gateway REST API implementation by injecting code that reads explicitly from the form-data and embeds it in protoReq.Data. (This is the best option as it does not require custom handlers, however, it could be potentially unstable if the grpc-gateway changes its generator in the future)."

api/translate/buf.yaml
```diff
@@ -10,3 +10,4 @@ lint:
     - DEFAULT
   except:
     - ENUM_VALUE_PREFIX
+    - RPC_RESPONSE_STANDARD_NAME
```
VladislavsPerkanuks, line 13: was contradicting with google AIP

VladislavsPerkanuks:
Need a little help with tests, sometimes `resp, err := client.Do(req)` returns `nil, nil`, which causes panic when trying assert. No idea why

.vscode/settings.json
```diff
@@ -0,0 +1,15 @@
+{
+  "emeraldwalk.runonsave": {
+    "autoClearConsole": true,
+    "commands": [
+      {
+        "match": "\\.proto$",
+        "cmd": "earthly +proto"
+      },
+      {
+        "match": "\\.mod$",
```
janishorsts, line 10: Interesting.

```suggestion
        "match": "^go.mod$",
```

cmd/translate/main_grpc_integration_test.go
```diff
@@ -0,0 +1,141 @@
...
+			want: codes.InvalidArgument,
+		},
+	}
+
+	for _, tt := range tests {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+
+			conn := createConnection(ctx, t)
```
klavinsjanis, line 132: Do you have to create a connection for each test?

cmd/translate/main_grpc_integration_test.go
```diff
@@ -0,0 +1,141 @@
...
+	os.Exit(m.Run())
+}
+
+func createConnection(ctx context.Context, t *testing.T) *grpc.ClientConn {
+	t.Helper()
+
+	conn, err := grpc.DialContext(
+		ctx,
+		"localhost:8080",
+		grpc.WithTransportCredentials(insecure.NewCredentials()),
```
klavinsjanis, line 33: Wouldn't it be better to find a free port? Port 8080 might not always be available.

cmd/translate/main_grpc_integration_test.go
VladislavsPerkanuks: My aim intention to increase its independence.
Each request is initiated by a new connection and client, but I'm not sure if that's more accurate or superior to having a single client and connection to test all endpoints.

cmd/translate/main_grpc_integration_test.go
```diff
@@ -0,0 +1,141 @@
...
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/credentials/insecure"
+	"google.golang.org/grpc/status"
+)
+
+func TestMain(m *testing.M) {
+	go main()
+
+	// Needed to prevent panic, when service is not ready to receive requests
+	time.Sleep(time.Millisecond * 50)
```
janishorsts, line 22: Do your best to avoid using **time.Sleep()**:
- it slows down tests
- it is not deterministic

One of the simplest ways to resolve it is by checking if the port is open https://stackoverflow.com/questions/56336168/golang-check-tcp-port-open .

cmd/translate/main_grpc_integration_test.go
```diff
@@ -0,0 +1,141 @@
...
+func Test_UploadTranslationFile_gRPC(t *testing.T) {
+	t.Parallel()
+
+	ctx := context.Background()
+
+	type args struct {
+		req *pb.UploadTranslationFileRequest
+	}
+
+	tests := []struct {
```
janishorsts, line 51: We'll do a workshop on testing. To make sure we write tests efficiently with the limited resources we have. The complexity of testing snowballs the higher we are in the test pyramid.

cmd/translate/main_rest_integration_test.go
```diff
@@ -0,0 +1,133 @@
...
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+
+			body, contentType := attachFile(tt.args.text, t)
+
+			req, err := http.NewRequestWithContext(ctx, "PUT", fmt.Sprintf("%s/%s", addr, tt.args.language), body)
+			assert.NoError(t, err)
+
+			req.Header.Add("Content-Type", contentType)
+			client := &http.Client{}
```
janishorsts, line 124: another way of declaring variable of *T with zero values

```suggestion
			client := new(http.Client)
```

cmd/translate/main_rest_integration_test.go
```diff
@@ -0,0 +1,133 @@
...
+			resp, err := client.Do(req)
+			if err == nil {
```
janishorsts, line 126: always check err

```suggestion
			if !assert.NoError(t, err) {
				return
			}
			
			defer resp.Body.Close()
```

janishorsts (changes requested):

cmd/translate/main_grpc_integration_test.go
VladislavsPerkanuks: I believe this is not a priority at the momennt

cmd/translate/main_rest_integration_test.go
VladislavsPerkanuks: Noted, but it has zero difference with `&T{}`. And this syntax is more explicit, so I prefer it over `new()`

.vscode/settings.json
VladislavsPerkanuks: `^` wildcard is not working in this extension for some reason, so closest I can get is `go\\.mod$` which will match `*go.mod`, I guess it good enough

klavinsjanis (approved):

cmd/translate/main_grpc_integration_test.go
```diff
@@ -0,0 +1,130 @@
...
+func createConnection(ctx context.Context, t *testing.T) *grpc.ClientConn {
+	t.Helper()
+
+	conn, err := grpc.DialContext(
+		ctx,
+		"localhost:8080",
+		grpc.WithTransportCredentials(insecure.NewCredentials()),
+	)
+	if err != nil {
+		log.Panic("Failed to dial bufnet: %w", err)
```
janishorsts, line 25: What is bufnet? Simply, "connect to translate service".

cmd/translate/main_rest_integration_test.go
```diff
@@ -0,0 +1,152 @@
...
+
+	body := &bytes.Buffer{}
+
+	writer := multipart.NewWriter(body)
+	defer writer.Close()
+
+	part, err := writer.CreateFormFile("file", "test.json")
+	if err != nil {
+		log.Panic(err)
+	}
```
janishorsts, line 43: Inconsistent style with other tests.

```suggestion
	asert.NoError(t, err)
	
	// or
	
	if !assert.NoError(t, err) {
		return nil, ""
	}
```

janishorsts (changes requested):

janishorsts (approved):

PR #16, opened 2023-02-13
#15 Transform to and from ngx translate

DarjaLeonova:


pkg/convert/json_ngxtranslate.go
```diff
@@ -0,0 +1,61 @@
...
+	"fmt"
+
+	"go.expect.digital/translate/pkg/model"
+)
+
+func FromNgxTranslate(b []byte) (messages model.Messages, err error) {
+	var dst map[string]interface{}
+
+	if err = json.Unmarshal(b, &dst); err != nil {
+		return messages, fmt.Errorf("unmarshal ngx translateOld: %w", err)
```
janishorsts, line 15: translateOld?

pkg/convert/json_ngxtranslate.go
```diff
@@ -0,0 +1,61 @@
...
+					return err
+				}
+			}
+		}
+
+		return err
+	}
+
+	if err = traverseMap("", dst); err != nil {
+		return messages, fmt.Errorf("traverse ngx: %w", err)
```
janishorsts, line 42: ngx? Full name is `ngx-translate`.

pkg/convert/json_ngxtranslate.go
```diff
@@ -0,0 +1,61 @@
...
+func ToNgxTranslate(messages model.Messages) (b []byte, err error) {
+	dst := make(map[string]string, len(messages.Messages))
+
+	for _, msg := range messages.Messages {
+		dst[msg.ID] = msg.Message
+	}
+
+	b, err = json.Marshal(dst)
+	if err != nil {
+		return nil, fmt.Errorf("marshal to ngx translateOld : %w", err)
```
janishorsts, line 57: translateOld?

pkg/convert/json_ngxtranslate_test.go
```diff
@@ -0,0 +1,129 @@
...
+					},
+				},
+			},
+			wantErr: false,
+		},
+		{
+			name:       "Nested mixed",
+			serialized: []byte(`{"message.example":"message1","message":{"example":"message1"}}`),
+			expected: model.Messages{
+				Messages: []model.Message{
```
janishorsts, line 62: Identifiers are unique. It should not be possible to have two messages with the same ID.

janishorsts (changes requested):

pkg/convert/json_ngxtranslate_test.go
```diff
@@ -0,0 +1,129 @@
...
+)
+
+func Test_fromNgxTranslate(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name       string
+		serialized []byte
+		expected   model.Messages
+		wantErr    bool
```
VladislavsPerkanuks, line 17: Unused

pkg/convert/json_ngxtranslate_test.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+func Test_toNgxTranslate(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name     string
+		messages model.Messages
+		expected []byte
+		wantErr  bool
```
VladislavsPerkanuks, line 97: Unused

pkg/convert/json_ngxtranslate_test.go
```diff
@@ -0,0 +1,129 @@
...
+		},
+	}
+	for _, tt := range tests {
+		tt := tt
+
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+
+			result, err := ToNgxTranslate(tt.messages)
+			assert.NoError(t, err)
```
VladislavsPerkanuks, line 124:
```suggestion
			if !assert.NoError(t, err) {
				return
			}
```
1 failed assertion instead of 2, when `err!=nil`

pkg/convert/json_ngxtranslate_test.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+
+	for _, tt := range tests {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+
+			result, err := FromNgxTranslate(tt.serialized)
+
+			assert.NoError(t, err)
```
VladislavsPerkanuks, line 84:
```suggestion
			if !assert.NoError(t, err) {
				return
			}
```
1 failed assertion instead of 2, when `err!=nil`

pkg/convert/json_ngxtranslate.go
```diff
@@ -0,0 +1,61 @@
...
+	if err = json.Unmarshal(b, &dst); err != nil {
+		return messages, fmt.Errorf("unmarshal from ngx-translate to model.Messages: %w", err)
+	}
+
+	var traverseMap func(key string, value interface{}) error
+
+	traverseMap = func(key string, value interface{}) (err error) {
+		switch v := value.(type) {
+		default:
+			return errors.New("wrong key type")
```
klavinsjanis, line 23: Unsupported value type?

VladislavsPerkanuks: Not covered in tests, but could be

DarjaLeonova: Thank you ;)

pkg/convert/json_ngxtranslate.go
```diff
@@ -0,0 +1,61 @@
...
+func ToNgxTranslate(messages model.Messages) (b []byte, err error) {
+	dst := make(map[string]string, len(messages.Messages))
+
+	for _, msg := range messages.Messages {
+		dst[msg.ID] = msg.Message
+	}
+
+	b, err = json.Marshal(dst)
+	if err != nil {
+		return nil, fmt.Errorf("marshal to ngx-translate from model.Messages : %w", err)
```
klavinsjanis, line 57: Messages: %w

pkg/convert/json_ngxtranslate.go
VladislavsPerkanuks: I would use something more explicit, like "wrong key type '{keyType}' expected string/map[string]interface{}"

pkg/convert/json_ngxtranslate.go
```diff
@@ -0,0 +1,61 @@
...
+	if err = json.Unmarshal(b, &dst); err != nil {
+		return messages, fmt.Errorf("unmarshal from ngx-translate to model.Messages: %w", err)
+	}
+
+	var traverseMap func(key string, value interface{}) error
+
+	traverseMap = func(key string, value interface{}) (err error) {
+		switch v := value.(type) {
+		default:
+			return errors.New("unsupported value type")
```
klavinsjanis, line 23: This error might be better with some added details: `fmt.Errorf("usupported value type %T for key %s", value, key)`

pkg/convert/json_ngxtranslate_test.go
DarjaLeonova: What do you mean by unused? We have more than one test so we have a struct array and we are using mentioned properties in for loop to run all tests

VladislavsPerkanuks: `wantErr` is defined in struct, but unused in actual tests

pkg/convert/json_ngxtranslate.go
DarjaLeonova: Thank you ;)

pkg/convert/json_ngxtranslate.go
```diff
@@ -0,0 +1,60 @@
...
+	if err = json.Unmarshal(b, &dst); err != nil {
+		return messages, fmt.Errorf("unmarshal from ngx-translate to model.Messages: %w", err)
+	}
+
+	var traverseMap func(key string, value interface{}) error
+
+	traverseMap = func(key string, value interface{}) (err error) {
+		switch v := value.(type) {
+		default:
+			return fmt.Errorf("usupported value type %T for key %s", value, key)
```
janishorsts, line 22:
```suggestion
			return fmt.Errorf("unsupported value type %T for key %s", value, key)
```

janishorsts (changes requested):

PR #20, opened 2023-02-13
17-transform-to-and-from-arb

VladislavsPerkanuks:


pkg/convert/arb.go
```diff
@@ -0,0 +1,114 @@
...
+		description, err := findDescription(key)
+		if err != nil {
+			return model.Messages{}, fmt.Errorf("find description of '%s': %w", key, err)
+		}
+
+		messages.Messages = append(messages.Messages, model.Message{
+			ID:          key,
+			Message:     value,
+			Description: description,
+		})
```
janishorsts, line 91: The following reduces code boilerplate by reducing the number of variables.

```suggestion
		msg  := messages.Message{ID: key}
		
		msg.Message, ok = value.(string)
		if !ok {
			return model.Messages{}, fmt.Errorf("unsupported value type '%T' for key '%s'", value, key)
		}

		msg.Description, err = findDescription(key)
		if err != nil {
			return model.Messages{}, fmt.Errorf("find description of '%s': %w", key, err)
		}

		messages.Messages = append(messages.Messages, msg)
```

pkg/convert/arb.go
```diff
@@ -0,0 +1,114 @@
...
+	return messages, nil
+}
+
+// Converts model.Messages into a serialized data in ARB file format.
+func ToArb(messages model.Messages) ([]byte, error) {
+	dst := make(map[string]interface{}, len(messages.Messages))
+
+	for _, msg := range messages.Messages {
+		dst[msg.ID] = msg.Message
+		if len(msg.Description) != 0 {
```
janishorsts, line 103:
```suggestion
		// the intent is more precise with this one
		if len(msg.Description) > 0 {
		
		// or

		If msg.Description != "" {
```

pkg/convert/arb_test.go
```diff
@@ -0,0 +1,150 @@
...
+)
+
+func Test_FromArb(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name    string
+		data    []byte
+		want    model.Messages
+		wantErr bool
```
janishorsts, line 17:
```suggestion
		wantErr error
```

janishorsts (changes requested):

pkg/convert/arb.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+
+	findDescription := func(key string) (string, error) {
+		// if '@' prefix missing then no additional information is provided
+		subKey, ok := dst["@"+key]
+		if !ok {
+			return "", nil
+		}
+
+		keyMap, ok := subKey.(map[string]interface{})
```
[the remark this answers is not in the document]
VladislavsPerkanuks: I don't quite get what you meant here.
All information I got is from [flutter documentation](https://docs.flutter.dev/development/accessibility-and-localization/internationalization)

If I understand correctly, concrete type means that type has a full definition, which in this case is true. if there is a key for example **greeting** and there is metadata key \@greeting, which contains additional information for the message, such as Description, Placeholders, etc., then \@greeting has a defined structure, as mentioned in the comment
```json
"greeting": "Welcome {user}!",
"@greeting": {
  "description":"Used to greet specific user to website",
  "placeholders": {
    "user":{
      "type":"string",
      "example":"Bob"
      }
    }
  }
```

The metadata key can contain additional metadata keys, but **description** and **placeholders** are the most used ones.

pkg/convert/arb.go
```diff
@@ -0,0 +1,107 @@
...
+
+		messages.Messages = append(messages.Messages, msg)
+	}
+
+	return messages, nil
+}
+
+// Converts model.Messages into a serialized data in ARB file format.
+func ToArb(messages model.Messages) ([]byte, error) {
+	dst := make(map[string]interface{}, len(messages.Messages)*2) //nolint:gomnd
```
VladislavsPerkanuks, line 92: Due to additional metadata keys, the map size is at least `len(messages.Messages)` and the max size of a map can be up to `len(messages.Messages)*2`.
It is better to allocate a max size beforehand, afaik

PR #22, opened 2023-02-15
21-download-file

VladislavsPerkanuks:


pkg/translate/translate.go
```diff
@@ -14,34 +15,89 @@ type TranslateServiceServer struct {
...
 	ctx context.Context,
 	req *pb.UploadTranslationFileRequest,
 ) (*emptypb.Empty, error) {
-	var (
-		reqLanguage = req.GetLanguage()
-		reqData     = req.GetData()
-		reqSchema   = req.GetSchema()
-	)
+	reqParams := UploadParams{
+		Language: LanguageData{Str: req.GetLanguage()},
```
janishorsts, line 49: There are at least two steps involved in processing a request:

- parse request (type/schema guard)
- validate request (logical guard)
- access control
- etc

If you separate these steps, it simplifies the flow and removes unnecessary temporary data (E.g. Language struct with Str and Tag fields) that leaks and generates noise.

```go
// one of the options is...

type uploadParams struct {
  Language language.Tag
  // ...
}

params, err := parseUploadParams(req *pb.UploadTranslationFileRequest) (*uploadParams, error)

// ...

if err := params.validate(); err != nil {
```

janishorsts (changes requested):

pkg/translate/translate.go
```diff
@@ -14,34 +15,89 @@ type TranslateServiceServer struct {
 	pb.UnimplementedTranslateServiceServer
 }
 
+// ----------------------UploadTranslationFile-------------------------------
+
+type UploadParams struct {
```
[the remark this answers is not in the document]
VladislavsPerkanuks: Yeah, I wanted to do that but ran into problems when was making integration tests specifically for REST. My idea was to use the same `translate.UploadParams` to create table tests.
```go
	tests := []struct {
		name   string
		params translate.UploadParams
		want   uint
	}
```

What is the preferred solution then, if we make `uploadParams` inaccessible for integration tests? Creating custom struct for tests?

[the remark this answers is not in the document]
janishorsts: yes, define a new struct in tests

klavinsjanis (approved):

pkg/translate/translate.go
```diff
@@ -14,34 +15,94 @@ type TranslateServiceServer struct {
...
+
+// Validates request parameters for UploadTranslationFile.
+func (u *uploadParams) validate() error {
+	if len(u.Data) == 0 {
+		return fmt.Errorf("'data' is required")
+	}
+
+	// Enforce that schema is present. (Temporal solution)
+	if u.Schema == pb.Schema_UNSPECIFIED {
+		return fmt.Errorf("'schema' must be specified")
```
janishorsts, line 51:
```suggestion
		return fmt.Errorf("'schema' is required")
```

pkg/translate/translate.go
```diff
@@ -14,34 +15,94 @@ type TranslateServiceServer struct {
...
+	Data     []byte
+	Schema   pb.Schema
+}
+
+func parseUploadParams(req *pb.UploadTranslationFileRequest) (*uploadParams, error) {
+	reqLanguage := req.GetLanguage()
+
+	tag, err := language.Parse(reqLanguage)
+	if err != nil {
+		return nil, fmt.Errorf("parsing language '%s': %w", reqLanguage, err)
```
janishorsts, line 31: shorter, the verb in simple present form

```suggestion
		return nil, fmt.Errorf("parse language '%s': %w", reqLanguage, err)
```

janishorsts (approved):

PR #24, opened 2023-02-17
Instrument traces (OpenTelemetry)

VladislavsPerkanuks:
#23


pkg/translate/translate.go
```diff
@@ -97,6 +101,9 @@ func (t *TranslateServiceServer) DownloadTranslationFile(
 	ctx context.Context,
 	req *pb.DownloadTranslationFileRequest,
 ) (*pb.DownloadTranslationFileResponse, error) {
+	span := trace.SpanFromContext(ctx)
+	defer span.End()
```
VladislavsPerkanuks, line 105: The result looks like this:
image

pkg/translate/translate_test.go
```diff
@@ -96,7 +96,7 @@ func Test_ValidateUploadParams(t *testing.T) {
 				Data:     []byte(`{"key":"value"}`),
 				Schema:   pb.Schema_UNSPECIFIED,
 			},
-			wantErr: errors.New("'schema' must be specified"),
+			wantErr: errors.New("'schema' is required"),
```
VladislavsPerkanuks, line 99: Overlooked this in the previous merge

pkg/translate/translate.go
VladislavsPerkanuks: Modified context is accessible via `trace.ContextWithSpan` e.g. 
```go
newCtx := trace.ContextWithSpan(ctx,span)
```

cmd/translate/main.go
```diff
@@ -9,29 +9,46 @@ import (
...
 // rootCmd represents the base command when called without any subcommands.
 var rootCmd = &cobra.Command{
 	Use:   "translate",
 	Short: "Enables translation for Cloud-native systems",
 	Long:  `Enables translation for Cloud-native systems`,
 	Run: func(cmd *cobra.Command, args []string) {
-		grpcSever := grpc.NewServer()
+		var err error
+
+		tp, err = tracer.TracerProvider("http://localhost:14268/api/traces", "translate")
```
klavinsjanis, line 38: Perhaps would be better to store these URLs in a .env file?

pkg/translate/translate.go
```diff
@@ -58,6 +59,9 @@ func (t *TranslateServiceServer) UploadTranslationFile(
 	ctx context.Context,
 	req *pb.UploadTranslationFileRequest,
 ) (*emptypb.Empty, error) {
+	span := trace.SpanFromContext(ctx)
```
klavinsjanis, line 62: Do spans have to be added manually? There weren't any OTEL wrappers for this purpose?

cmd/translate/main_rest_integration_test.go
```diff
@@ -28,7 +28,15 @@ func TestMain(m *testing.M) {
 
 	conn.Close()
 
-	os.Exit(m.Run())
+	code := m.Run()
+
+	// To avoid tp.ForceFlush() from indefinitely blocking, allocate 5 seconds for exporting spans.
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+
+	tp.ForceFlush(ctx)
```
klavinsjanis, line 36: ForceFlush might not be necessary after tp shutdown is added.

cmd/translate/main.go
```diff
@@ -9,29 +9,46 @@ import (
...
+		grpcSever := grpc.NewServer(
+			grpc.UnaryInterceptor(otelgrpc.UnaryServerInterceptor()),
+			grpc.StreamInterceptor(otelgrpc.StreamServerInterceptor()),
+		)
+
 		mux := runtime.NewServeMux()
 		pb.RegisterTranslateServiceServer(grpcSever, &translate.TranslateServiceServer{})
-		err := pb.RegisterTranslateServiceHandlerFromEndpoint(
+
+		err = pb.RegisterTranslateServiceHandlerFromEndpoint(
```
janishorsts, line 51: check error

cmd/translate/main.go
```diff
@@ -57,18 +74,18 @@ var rootCmd = &cobra.Command{
 		grpcL := m.Match(cmux.HTTP2())
 
 		go func() {
-			if err := server.Serve(httpL); err != nil {
+			if err = server.Serve(httpL); err != nil {
```
janishorsts, line 77: You write to the same error instance from the main thread from two go routines, avoid it, it may cause panic. Use a local variable.

```suggestion
			if err := server.Serve(httpL); err != nil {
```

cmd/translate/main.go
janishorsts: @VladislavsPerkanuks please a new ticket to discuss the implementation and naming convention for variables in the configuration file, environment variables and CLI arguments.

cmd/translate/main_rest_integration_test.go
```diff
@@ -28,7 +28,15 @@ func TestMain(m *testing.M) {
 
 	conn.Close()
 
-	os.Exit(m.Run())
+	code := m.Run()
+
+	// To avoid tp.ForceFlush() from indefinitely blocking, allocate 5 seconds for exporting spans.
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
```
janishorsts, line 34: Do not use timeout context, tests already have a timeout (30 seconds by default). Use **context.Background()**.

pkg/tracer/tracer.go
```diff
@@ -0,0 +1,35 @@
...
+	"fmt"
+
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/exporters/jaeger"
+	"go.opentelemetry.io/otel/sdk/resource"
+	tracesdk "go.opentelemetry.io/otel/sdk/trace"
+	semconv "go.opentelemetry.io/otel/semconv/v1.4.0"
+)
+
+func TracerProvider(url, service string) (*tracesdk.TracerProvider, error) {
```
janishorsts, line 13: Does OTEL provide a default TracerProvider based on the environment variables? https://opentelemetry.io/docs/reference/specification/sdk-environment-variables/

pkg/translate/translate.go
janishorsts: No, it's not needed - the GRPC server should have one. https://github.com/translate-agent/translate/pull/24/files#diff-06a04b9bbf6fc3785a7a7389f59f4cf08c4133c69c665dd689072445a43d632aR44

cmd/translate/main.go
VladislavsPerkanuks: Is already checked in
https://github.com/translate-agent/translate/blob/eb1cd85852fc834c86b8a44886c47fc618dae38e/cmd/translate/main.go#L56-L58

pkg/translate/translate.go
VladislavsPerkanuks: You are right

pkg/tracer/tracer.go
```diff
@@ -0,0 +1,33 @@
...
+
+import (
+	"context"
+	"fmt"
+
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracehttp"
+	"go.opentelemetry.io/otel/sdk/resource"
+	"go.opentelemetry.io/otel/sdk/trace"
+	tracesdk "go.opentelemetry.io/otel/sdk/trace"
```
janishorsts, line 11: Package `trace` imported twice.

pkg/tracer/tracer.go
```diff
@@ -0,0 +1,33 @@
...
+	semconv "go.opentelemetry.io/otel/semconv/v1.4.0"
+)
+
+func TracerProvider() (func(context.Context) error, error) {
```
janishorsts, line 15:
```suggestion
func TracerProvider() (*trace.TracerProvider, error) {
```

janishorsts (changes requested):

.gitignore
```diff
@@ -10,6 +10,8 @@
 
 # Test binary, built with `go test -c`
 *.test
```
janishorsts, line 12: All binaries are built in `/bin`

cmd/translate/main.go
```diff
@@ -84,6 +125,7 @@ func main() {
 func init() {
 	cobra.OnInitialize(initConfig)
 	rootCmd.PersistentFlags().StringVar(&cfgFile, "config", "", "config file (default is ./translate.yaml)")
+	rootCmd.PersistentFlags().StringVar(&envFile, "env", "", "environment file")
```
janishorsts, line 128: We can remove this flag, we'll never load env variable files in the **translate** service.

There are other ways to consume env variable files. E.g. when running tests in the VS Code, we can let go extension know where to find the env file with configuration `"go.testEnvFile": "${workspaceFolder}/.env.test"`. And so on.

.gitignore
VladislavsPerkanuks: that was default .gitignore for `go` application, but oki

cmd/translate/main_rest_integration_test.go
```diff
@@ -18,17 +20,35 @@ import (
 const baseAddr = "http://localhost:8080"
 
 func TestMain(m *testing.M) {
-	go main()
+	envFile = "../../.env.test"
```
janishorsts, line 23: `go.testEnvFile` does load env file

.gitignore
janishorsts: We have to be organized and keep it tidy. Creating binaries everywhere has side effects, e.g the cache determination in the earthly when copying directories with `COPY --dir` and expecting only go files.

What is the use case for running `go test -c`? Why would you use it?

VladislavsPerkanuks: As I said, this is template from https://github.com/github/gitignore
[go .gitignore](https://github.com/github/gitignore/blob/main/Go.gitignore) 
There is also [vscode extension](https://marketplace.visualstudio.com/items?itemName=codezombiech.gitignore) that I used : 

I simply did not argue with it contents and presumed that everything is good there

Issue #19, opened 2023-02-13
Transform to and from POT


DarjaLeonova:
https://www.gnu.org/software/gettext/manual/html_node/PO-Files.html

PR #26, opened 2023-02-23
#19 Transform-to-and-from-POT

DarjaLeonova:
Created a parser for our model.Messages struct. The change consists of 3 parts: lexer, AST, and Pot.
1. Lexer - reads the PO format file and divides its entries to []Tokens. (We support most used PO entries). Also, we support multiline for the following token types: MsgId, MsgStr, PluralMsgId, and PluralMsgStr.
2.  AST - takes our []Tokens and validates tokens content, filters them, and collects them into a single structure (PO struct).
3. Pot - Transforms from []byte to our model.Messages (based on AST) and from the model.Messages to []byte. (Note most of the PO tokens are not used in our models.Messages so we simply ignore them)


pkg/convert/pot.go
```diff
@@ -0,0 +1,70 @@
...
+		b.WriteString(fmt.Sprintf("msgstr \"%s\"\n", message.Message))
+	}
+
+	return b.Bytes(), nil
+}
+
+func FromPot(b []byte) (model.Messages, error) {
+	var po *pot.Po
+	if err := json.Unmarshal(b, &po); err != nil {
+		return model.Messages{}, fmt.Errorf("error unmarshaling byte slice to pot.Po: %w", err)
```
VladislavsPerkanuks, line 38:
```suggestion
  return model.Messages{}, fmt.Errorf("unmarshal byte slice to pot.Po: %w", err)
```

pkg/convert/pot.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+
+	messages := make([]model.Message, 0, len(po.Messages))
+
+	for _, node := range po.Messages {
+		if node.MsgIdPlural != "" {
```
VladislavsPerkanuks, line 44: I am curious why are we skipping plurals ?

DarjaLeonova: Because in our model.Messages structure we don't support plurals. I had conversation with @janishorsts about this ;)

pkg/convert/pot.go
```diff
@@ -0,0 +1,70 @@
...
+			continue
+		}
+
+		var fuzzy bool
+		if strings.Contains(node.Flag, "fuzzy") {
+			fuzzy = true
+		}
```
VladislavsPerkanuks, line 51:
```suggestion
fuzzy := strings.Contains(node.Flag, "fuzzy")
```
No redundant declaration + cleaner

pkg/convert/pot.go
```diff
@@ -0,0 +1,70 @@
...
+)
+
+func ToPot(m model.Messages) ([]byte, error) {
+	var b bytes.Buffer
+
+	b.WriteString(fmt.Sprintf("\"Language: %s\n", m.Language))
+
+	for _, message := range m.Messages {
+		if message.Description != "" {
+			b.WriteString(fmt.Sprintf("#. %s\n", message.Description))
```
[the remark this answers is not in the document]
VladislavsPerkanuks: Either way, error should be checked

pkg/pot/lexer.go
```diff
@@ -0,0 +1,132 @@
...
+
+		line = strings.TrimSpace(line)
+		if len(line) == 0 {
+			continue
+		}
+
+		switch {
+		case strings.HasPrefix(line, "# Language:"):
+			tokens = append(tokens, Token{Value: parseMsgString(line), Type: HeaderLanguage})
+			continue
```
VladislavsPerkanuks, line 52: redundant `continue`
no need switch executes only once by default

pkg/pot/lexer.go
```diff
@@ -0,0 +1,132 @@
...
+	MsgidPrevUntdStr
+)
+
+type Token struct {
+	Value string
+	Type  TokenType
+	Index int // plural index for msgstr with plural forms
+}
+
+func Lex(r *bufio.Reader) []Token {
```
VladislavsPerkanuks, line 35: Maybe I don't understand but this function is used only in tests, why is it needed?

pkg/pot/lexer.go
```diff
@@ -0,0 +1,132 @@
...
+	var tokens []Token
+
+	for {
+		line, err := r.ReadString('\n')
+		if err != nil {
+			break
```
VladislavsPerkanuks, line 41:
```suggestion
		if err == io.EOF {
			break
		} else if err != nil {
			panic(err) // panic or return err, dunno about that
		}
```

pkg/pot/lexer.go
```diff
@@ -0,0 +1,132 @@
...
+			tokens = append(tokens, Token{Value: parseMsgString(line), Type: MsgctxtPreviousContext})
+			continue
+		case strings.HasPrefix(line, "#,"):
+			tokens = append(tokens, Token{Value: parseMsgString(line), Type: Flag})
+			continue
+		case strings.HasPrefix(line, "#"):
+			tokens = append(tokens, Token{Value: parseMsgString(line), Type: TranslatorComment})
+			continue
+		default:
+			if strings.HasPrefix(line, `"`) {
```
VladislavsPerkanuks, line 106: choose one, otherwise looks not right
1. `default` without ``if strings.HasPrefix(line, `"`)``
2.  without `default` with ``case strings.HasPrefix(line, `"`):``

pkg/pot/lexer.go
```diff
@@ -0,0 +1,132 @@
...
+				tokenToAppend := tokens[len(tokens)-1]
+				if tokenToAppend.Type == MsgId || tokenToAppend.Type == MsgStr ||
+					tokenToAppend.Type == PluralMsgId || tokenToAppend.Type == PluralMsgStr {
```
VladislavsPerkanuks, line 109:
```suggestion
				switch tokenToAppend.Type {
				case MsgId, MsgStr, PluralMsgId, PluralMsgStr:
				    ...
				}
```

pkg/convert/pot.go
```diff
@@ -0,0 +1,70 @@
...
+		}
+
+		if message.Fuzzy {
+			b.WriteString("#, fuzzy\n")
+		}
+
+		b.WriteString(fmt.Sprintf("msgid \"%s\"\n", message.ID))
+
+		b.WriteString(fmt.Sprintf("msgstr \"%s\"\n", message.Message))
```
janishorsts, line 29: - What happens if `message.Message` value is `Hello "World"`? Currently, it writes `msgstr "Hello "World""` to POT. How to escape double quotes in POT?
- What happens if `message.Message` contains a multiline value?

pkg/convert/pot.go
janishorsts: https://github.com/orgs/translate-agent/projects/1/views/1?pane=issue&itemId=21251425

pkg/convert/pot.go
```diff
@@ -0,0 +1,97 @@
...
+	"strings"
+
+	"go.expect.digital/translate/pkg/model"
+	"go.expect.digital/translate/pkg/pot"
+)
+
+func ToPot(m model.Messages) ([]byte, error) {
+	var b bytes.Buffer
+
+	b.WriteString(fmt.Sprintf("\"Language: %s\n", m.Language))
```
janishorsts, line 16: check err

pkg/convert/pot.go
```diff
@@ -0,0 +1,97 @@
...
+		}
+
+		if message.Fuzzy {
+			_, err := fmt.Fprint(&b, "#, fuzzy\n")
+			if err != nil {
+				return nil, fmt.Errorf("write fuzzy: %w", err)
+			}
+		}
+
+		if strings.HasSuffix(message.ID, "\\n") {
```
janishorsts, line 33: can id really be multiline?

janishorsts: I guess, it can

DarjaLeonova: Yes, msgid in a PO file can be a multiline string
image

pkg/convert/pot.go
```diff
@@ -0,0 +1,97 @@
...
+			_, err := fmt.Fprintf(&b, "msgid \"%s\"\n", messageIdWithQuotes)
+			if err != nil {
+				return nil, fmt.Errorf("write msgid: %w", err)
+			}
+		}
+
+		if strings.HasSuffix(message.Message, "\\n") {
+			_, err := fmt.Fprintf(&b, "msgstr \"\" \n\"%s\"\n", message.Message)
+			if err != nil {
+				return nil, fmt.Errorf("write msgid: %w", err)
```
janishorsts, line 49:
```suggestion
				return nil, fmt.Errorf("write msgstr: %w", err)
```

pkg/convert/pot.go
```diff
@@ -0,0 +1,97 @@
...
+			}
+		} else {
+			messageWithQuotes := strings.ReplaceAll(message.Message, "\"", "\\\"")
+			_, err := fmt.Fprintf(&b, "msgstr \"%s\"\n", messageWithQuotes)
+			if err != nil {
+				return nil, fmt.Errorf("write msgid: %w", err)
```
janishorsts, line 55:
```suggestion
				return nil, fmt.Errorf("write msgstr: %w", err)
```

pkg/convert/pot.go
```diff
@@ -0,0 +1,97 @@
...
+
+	po, err := pot.TokensToPo(tokens)
+	if err != nil {
+		return model.Messages{}, fmt.Errorf("convert tokens to pot.Po: %w", err)
+	}
+
+	messages := make([]model.Message, 0, len(po.Messages))
+
+	for _, node := range po.Messages {
+		if node.MsgIdPlural != "" {
```
janishorsts, line 74: Please add the comment with reference to https://github.com/orgs/translate-agent/projects/1?pane=issue&itemId=21251425 that the implementation will come after the design decision.

pkg/convert/pot_test.go
```diff
@@ -0,0 +1,298 @@
...
+			},
+			expected: []byte(
+				"\"Language: en\n" +
+					"#. A simple greeting\n#, " +
+					"fuzzy\n" +
+					"msgid \"\" \n\"Hello, world!\\n\"\n\"very long string\\n\"\n" +
+					"msgstr \"Bonjour le monde!\"\n" +
+					"#. A farewell\n" +
+					"#, fuzzy\nmsgid \"Goodbye!\"\n" +
+					"msgstr \"Au revoir!\"\n"),
```
janishorsts, line 71: To improve readability, I suggest using multiline strings, which are much easier to read.

```suggestion
			expected: []byte(`Language: en
#. A simple greeting
#, fuzzy
msgid ""
....`),
```

pkg/convert/pot_test.go
```diff
@@ -0,0 +1,297 @@
+package convert
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"go.expect.digital/translate/pkg/model"
+	"golang.org/x/text/language"
+)
```
VladislavsPerkanuks, line 10: I suggest grouping these e.g.
```go
import (
  stdlib
  ...

  3rd party packages
  ...

  our packages
  ...
)
```
That way it is easier to follow imports
[jetbrains](https://stackoverflow.com/a/70668482/19159194)
[vscode](https://marketplace.visualstudio.com/items?itemName=aleksandra.go-group-imports)

pkg/convert/pot.go
VladislavsPerkanuks: inconsistent with other printing to buffer

pkg/convert/pot_test.go
janishorsts: It's ok to group in two groups as we do now - stdlib and others. Easier to maintain. I've seen the same grouping (3 groups) practice in other projects, and it is out of control - hard to maintain and developers don't care, it's chaos. I don't see real benefits.

VladislavsPerkanuks: Could you elaborate on the meaning of out of control ?

VladislavsPerkanuks: Because there is vscode extension and jetbrains config that automatically does that. IMO in that way it looks cleaner somehow

janishorsts: In general, there are no benefits, developers don't spend much time looking at the list of imported packages. It is easy to find imported packages anyway. It has not been enforced in other projects, and developers add imported packages to arbitrary groups.

I've used both approaches in different projects. The simplest one works the best, it is more consistent and easier for developers to maintain.

Before I gained my experience building projects in go, I would agree with your statement.

pkg/convert/pot_test.go
```diff
@@ -224,7 +243,8 @@ func Test_ToPot(t *testing.T) {
 			if !assert.NoError(t, err) {
 				return
 			}
-
+			fmt.Printf("exp: %v\n", string(tt.expected))
+			fmt.Printf("res: %v\n", string(result))
```
VladislavsPerkanuks, line 247: Remove print statements

pkg/convert/pot_test.go
```diff
@@ -0,0 +1,317 @@
...
+						Fuzzy:       true,
+					},
+				},
+			},
+			expected: []byte(`"Language: en
+#. A simple greeting
+#, fuzzy
+msgid "Hello, world!"
+msgstr "Bonjour le monde!"
+#. A farewell
```
janishorsts, line 44: Can we please add an empty line between messages to improve readability by human beings?

```suggestion
msgstr "Bonjour le monde!"

#. A farewell
```

pkg/convert/pot_test.go
```diff
@@ -0,0 +1,317 @@
...
+msgstr "Au revoir!"
+`),
+		},
+		{
+			name: "When msgid is multiline",
+			input: model.Messages{
+				Language: language.English,
+				Messages: []model.Message{
+					{
+						ID:          "Hello, world!\\n\"\n\"very long string\\n",
```
janishorsts, line 56: Is it supposed to be the following for msgid? 

```po
msgid "" 
"Hello, world!\n"
"very long string\n"
```

```suggestion
						ID:          `Hello, world!
very long string
`,
```

pkg/convert/pot_test.go
janishorsts: @DarjaLeonova do you print a string to see the string value of the byte slice?

janishorsts: Use `t.Log` to print log messages in the future. When running tests in parallel, `fmt` prints in the middle of all parallel tests running at the time. `t.Log` is attached to the test and is printed in the test output.

pkg/convert/pot_test.go
```diff
@@ -0,0 +1,317 @@
...
+			expectedErr: fmt.Errorf("convert tokens to pot.Po: invalid po file: no messages found"),
+		},
+	}
+
+	for _, tt := range tests {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			result, err := FromPot(tt.input)
+			if err != nil {
```
janishorsts, line 309: Do not use the actual value to drive assertions. ALWAYS use the expected values. This is a mistake made by so many developers. :)

In your case, if err is always nil, your assertion of `expected.Err` never takes place.

pkg/pot/ast.go
```diff
@@ -0,0 +1,143 @@
...
+	MsgCtxtPrevCtxt       string
+	MsgIdPrevUntPluralStr string
+	MsgIdPrevUnt          string
+	MsgStr                []string
+}
+
+type Po struct {
+	Header   *HeaderNode
+	Messages []*MessageNode
+}
```
janishorsts, line 39: No need for pointers.

```suggestion
type Po struct {
	Header   HeaderNode
	Messages []MessageNode
}
```

pkg/convert/pot_test.go
DarjaLeonova: If I correctly understood..then yes image

pkg/convert/pot_test.go
VladislavsPerkanuks: image

pkg/pot/ast.go
```diff
@@ -0,0 +1,143 @@
...
+	var messages []MessageNode
+
+	partsN := 2
+	currentMessage := MessageNode{}
+	header := HeaderNode{}
+
+	for _, token := range tokens {
+		switch token.Type {
+		case HeaderLanguage:
+			parts := strings.Split(token.Value, ": ")
```
janishorsts, line 51: I have a feeling POT allows optional whitespace. Does it?

```suggestion
			parts := strings.Split(token.Value, ":")
			// both work
			parts := strings.Split(token.Value, ": ")
```

pkg/convert/pot_test.go
```diff
@@ -306,9 +323,9 @@ func TestFromPot(t *testing.T) {
...
 			result, err := FromPot(tt.input)
-			if err != nil {
-				assert.Equal(t, tt.expectedErr, fmt.Errorf(err.Error()))
-				return
+
+			if tt.expectedErr != nil {
+				assert.Errorf(t, err, tt.expectedErr.Error())
 			}
 
 			assert.Equal(t, tt.expected, result)
```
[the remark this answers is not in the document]
DarjaLeonova: Thank you ;)

pkg/pot/ast.go
DarjaLeonova: Yes. you are right. PO files allow whitespaces in headers after the ":" character.

pkg/convert/pot_test.go
janishorsts: The original text does not contain `"`. We normalize the value

```pot
msgid ""
"Hello, world!\n"
"very long string\n"
```
to

```
Hello, world!
very long string
```

```go
x := "Hello, world!\nvery long string\n"
```

pkg/convert/pot.go
```diff
@@ -0,0 +1,143 @@
...
+	msgIdTextLines := strings.Split(string(msgIDBytes), "\\n")
+
+	if len(msgIdTextLines) == 1 {
+		return fmt.Sprintf("msgid %s\n", msgIdTextLines[0]), nil
+	}
+
+	lines := make([]string, 0, len(msgIdTextLines))
+
+	for i, line := range msgIdTextLines {
+		if len(msgIdTextLines)-1 == i && line == "\"" {
```
janishorsts, line 108: Can you please provide example when line is `\"`?

DarjaLeonova: For example when in the end of input we have "\n"

image

janishorsts: Trim quotes early, they are not required.

```
msgIDBytes = string(msgIDBytes[1:len(msgIDBytes)-1] // trim quotes
msgIdTextLines := strings.Split(string(msgIDBytes), "\\n")
```

pkg/convert/pot.go
```diff
@@ -0,0 +1,143 @@
...
+		}
+
+		line = strings.ReplaceAll(line, "\"", "")
+		lines = append(lines, "\""+line+"\\n\""+"\n")
+	}
+
+	return fmt.Sprintf("msgid \"\"\n%s", strings.Join(lines, "")), nil
+}
+
+func formatMsgStr(msgStr string) (string, error) {
```
janishorsts, line 119: formatMsgStr is pretty much identical to formatMsgId. The difference is keyword - msgid or msgstr.

pkg/convert/pot.go
DarjaLeonova: But the Po file requires qoutes otherwise Po format is not correct. Here in msgid is not in correct format then  
image

pkg/pot/lexer.go
```diff
@@ -0,0 +1,115 @@
...
+
+func Lex(r *bufio.Reader) []Token {
+	var tokens []Token
+
+	for {
+		line, err := r.ReadString('\n')
+		if err == io.EOF {
+			break
+		} else if err != nil {
+			panic(err)
```
VladislavsPerkanuks, line 44: I think we should return the error and not panic after all. 
It would be silly to terminate service if there is an error in the reading `.po` file.

VladislavsPerkanuks (approved):
I cannot comment on the logic as I haven't looked deeply into the structure of `.po` files, but otherwise looks good

pkg/pot/lexer.go
janishorsts: It's not gonna terminate the service, the server recovers from the panics and sends a trace to collectors. But you are right - return error.

pkg/convert/pot.go
```diff
@@ -0,0 +1,132 @@
...
+		msgStr, err := formatPoTags(message.Message, MsgStr)
+		if err != nil {
+			return nil, fmt.Errorf("format msgstr: %w", err)
+		}
+
+		if _, err = fmt.Fprint(&b, msgStr); err != nil {
+			return nil, fmt.Errorf("write msgstr: %w", err)
+		}
+
+		if _, err = fmt.Fprint(&b, "\n"); err != nil {
```
janishorsts, line 59: This always creates an empty line at the end of the file. It's better to add an empty line between messages.

```suggestion
	for i, message := range m.Messages {
		if i > 0 {
			if _, err = fmt.Fprint(&b, "\n"); err != nil {
```

pkg/convert/pot.go
```diff
@@ -0,0 +1,132 @@
...
+			return nil, fmt.Errorf("write new line: %w", err)
+		}
+	}
+
+	return b.Bytes(), nil
+}
+
+func FromPot(b []byte) (model.Messages, error) {
+	tokens, err := pot.Lex(bufio.NewReader(bytes.NewReader(b)))
```
janishorsts, line 68:
```suggestion
	tokens, err := pot.Lex(bytes.NewReader(b))
```

pkg/convert/pot.go
```diff
@@ -0,0 +1,132 @@
...
+		messages = append(messages, message)
+	}
+
+	return model.Messages{
+		Language: po.Header.Language,
+		Messages: messages,
+	}, nil
+}
+
+func formatPoTags(str string, poTag PoTag) (string, error) {
```
janishorsts, line 107: This is a little bit more efficient, pass `bytes.Buffer` and write to it directly.

```suggestion
func writeTag(b *bytes.Buffer, tag Tag, str string) (error) {
```

pkg/convert/pot.go
```diff
@@ -0,0 +1,132 @@
...
+	b, err := json.Marshal(str)
```
janishorsts, line 108:
```suggestion
	b, err := json.Marshal(str) // use JSON encoding to escape special characters
```

pkg/convert/pot.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+
+	multiline := make([]string, 0, len(lines))
+
+	for i, line := range lines {
+		if len(lines)-1 == i && line == "" {
+			continue
+		}
+
+		line = strings.ReplaceAll(line, "\"", "")
```
janishorsts, line 127: why do we do this?

pkg/convert/pot.go
```diff
@@ -0,0 +1,132 @@
...
+		multiline = append(multiline, "\""+line+"\\n\""+"\n")
```
janishorsts, line 128: Using backquotes simplifies it. I'd expect something like this...

```suggestion
		multiline = append(multiline, `"`+line+`\n"`)
```

pkg/convert/pot_test.go
```diff
@@ -0,0 +1,357 @@
...
+#, fuzzy
+msgid "Hello, world!"
+msgstr "Bonjour le monde!"
+
+#. A farewell
+#, fuzzy
+msgid "Goodbye!"
+msgstr "Au revoir!"
+
+`),
```
janishorsts, line 50:
```suggestion
msgstr "Au revoir!"
`),
```

pkg/pot/lexer.go
```diff
@@ -0,0 +1,116 @@
...
+			}
+		}
+	}
+
+	return tokens, nil
+}
+
+func parseMsgString(line string) string {
+	subStrN := 2
+	tokenValue := strings.TrimSpace(strings.SplitN(line, " ", subStrN)[1])
```
janishorsts, line 109: does it panic if there is no empty space?

pkg/pot/lexer.go
```diff
@@ -0,0 +1,116 @@
...
+
+	return strings.ReplaceAll(tokenValue, `"`, "")
+}
+
+func parseMultilineString(line string) string {
+	return strings.ReplaceAll(line, `"`, "")
```
janishorsts, line 115: What happens if text contains quotes? `msgid "The book title is \"Spade\"."`?

pkg/convert/pot.go
DarjaLeonova: By removing the double quotes from the line string, we can ensure that the output string is correctly formatted according to the Po file format.

pkg/pot/lexer.go
DarjaLeonova: The function does not panic if there are no empty spaces in the input string.

pkg/convert/pot.go
janishorsts: Give me example, please!

pkg/pot/lexer.go
DarjaLeonova: As I see lexer function doesn't support quoted string. But should it support it?

pkg/pot/lexer.go
janishorsts: If `line` does NOT contain space, it panics  - `strings.SplitN()` returns a string slice with one element and accessing element by index 1 panics. Can it happen when po file is broken, incorrectly formatted?

janishorsts: E.g. line like `msgstrooops`, `#.ok`.

pkg/convert/pot.go
DarjaLeonova: I checked and apparently, this line of code is unnecessary, because the input is:
ID: "Hello, world"
so it converts to msgid "Hello, world" correctly

pkg/pot/lexer.go
DarjaLeonova: I checked and it's panics, so you are right

pkg/pot/lexer.go
```diff
@@ -0,0 +1,121 @@
...
+
+		line = strings.TrimSpace(line)
+		if len(line) == 0 {
+			continue
+		}
+
+		line = strings.TrimSpace(line)
+		if len(line) == 0 {
+			continue
+		}
```
janishorsts, line 51: duplicate, remove

```suggestion
```

pkg/convert/pot.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+
+	multiline := make([]string, 0, len(lines))
+
+	for i, line := range lines {
+		if len(lines)-1 == i && line == "" {
+			continue
+		}
+
+		multiline = append(multiline, "\""+line+"\\n\""+"\n")
```
janishorsts, line 125: write to b directly, no need to create `multiline`

pkg/convert/pot.go
```diff
@@ -0,0 +1,134 @@
...
+			return fmt.Errorf("write %s: %w", tag, err)
+		}
+
+		return nil
+	}
+
+	multiline := make([]string, 0, len(lines))
+
+	for i, line := range lines {
+		if len(lines)-1 == i && line == "" {
```
janishorsts, line 121: Why does it matter if it is the last line? Is an empty line even possible?

pkg/convert/pot.go
```diff
@@ -0,0 +1,134 @@
...
+
+	if len(lines) == 1 {
+		_, err = fmt.Fprintf(b, "%s \"%s\"\n", tag, lines[0])
+		if err != nil {
+			return fmt.Errorf("write %s: %w", tag, err)
+		}
+
+		return nil
+	}
+
```
janishorsts, line 117: I think we can remove this `if` statement and use multiline `for` loop if...

```go
// at the beginning of the func
_, err := fmt.Fprintf(b, "%s ", tag)
```

pkg/convert/pot.go
janishorsts: Oh, for use cases when `"foobar\n"`. Got it.

janishorsts, line 121: Do it outside `for`.

```go
// Remove the empty string element. The line is empty when splitting by "\\n" and "\n" is the last character in str.
if lines[len(lines)-1] == "" {
  lines = lines[:len(lines)-1]
}
```

pkg/pot/lexer_test.go
```diff
@@ -0,0 +1,124 @@
...
+			input:    "",
+			expected: []Token(nil),
+		},
+		{
+			name: "When spaces are trimmed",
+			input: `
+
+`,
+			expected: []Token(nil),
+		},
```
janishorsts, line 107: No need to set zero values. Up to you.

``` suggestion
		{
			name:     "Empty input",
		},
		{
			name: "When spaces are trimmed",
			input: `

`,
		},
```

pkg/pot/lexer_test.go
```diff
@@ -0,0 +1,124 @@
+package pot
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestLex(t *testing.T) {
```
janishorsts, line 10: Add a test case for invalid pot as well. E.g.`msgstr"oops"`. Is it invalid by POT spec?

pkg/convert/pot.go
```diff
@@ -0,0 +1,144 @@
...
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	"go.expect.digital/translate/pkg/model"
+	"go.expect.digital/translate/pkg/pot"
+)
+
+type PoTag string
```
janishorsts, line 13: We expose only `func From<std>` and `func To<std>`.  Does `pot` package have `msgid` and `msgstr`?

pkg/convert/pot.go
```diff
@@ -0,0 +1,144 @@
...
+
+	for i, message := range m.Messages {
+		if i > 0 {
+			if _, err := fmt.Fprint(&b, "\n"); err != nil {
+				return nil, fmt.Errorf("write new line: %w", err)
+			}
+		}
+
+		if message.Description != "" {
+			if _, err := fmt.Fprintf(&b, "#. %s\n", message.Description); err != nil {
```
janishorsts, line 35: If we have a multiline description, doesn't it break **po**? Do we have to escape special characters in the description?

pkg/pot/lexer.go
```diff
@@ -0,0 +1,208 @@
...
+		line := scanner.Text()
+		if err := parseLine(line, &tokens); err != nil {
+			return nil, err
+		}
+	}
+
+	return tokens, nil
+}
+
+func parseLine(line string, tokens *[]Token) error {
```
janishorsts, line 71: It's better for `parseLine` to return Token - removes side effects and keeps it simple.

```suggestion
func parseLine(line string) (Token, error) {
```

pkg/pot/lexer.go
```diff
@@ -0,0 +1,208 @@
...
+	if err != nil {
+		return fmt.Errorf("incorrect format of header token: %w", err)
+	}
+
+	*tokens = append(*tokens, Token{Value: val, Type: tokenType})
+
+	return nil
+}
+
+func parseMsgToken(line string, tokenType tokenType, tokens *[]Token) error {
```
janishorsts, line 126: both parseMsgToken and parseHeaderToken look the same to me

janishorsts (approved):

Issue #25, opened 2023-02-21
Configuration file, environment and CLI argument implementation


VladislavsPerkanuks:
Implementation and naming convention for variables in the configuration file, environment variables and CLI arguments.

_Originally posted by @janishorsts in https://github.com/translate-agent/translate/pull/24#discussion_r1111668376_

janishorsts:
# Binding

In `viper`, the binding is done automatically by matching the names of the environment variables. Viper is case-insensitive.

```yaml
# yaml conf
port: 3000
```

```shell
# env variables
TRANSLATE_PORT=5000
```

```go
// go
viper.SetEnvPrefix("translate") // it'll read all environment variables starting with TRANSLATE
viper.AutomaticEnv()

viper.ReadInConfig() // reads from the configuration file

fmt.Println(viper.GetInt("port")) // prints 5000

// we can bind the env variable manually if required
viper.BindEnv("translate.service.port", "TRANSLATE_PORT")

// bind viper to pflags (cobra)
viper.BindPFlag("service.port", cmd.Lookup("port")))
```

_Can we bind the env variable `TRANSLATE_SERVICE_PORT` to the nested  variable `service.port` automatically?_ 

_How does the binding work with `viper.BindPFlags()`? Exact names? Can we specify the binding function?_ 

# Precedence

`Viper` and `pflags` already implement the order of precedence of reading values (higher priority first):
- CLI arguments (`pflags`)
- environment variables (`viper`)
- configuration file (`viper`)

VladislavsPerkanuks:
>Can we bind the env variable `TRANSLATE_SERVICE_PORT` to the nested variable `service.port` automatically?

Yes, using: 
```go
viper.BindPFlags(rootCmd.Flags())
... // Load config and/or .env
viper.SetEnvPrefix("translate")
viper.SetEnvKeyReplacer(strings.NewReplacer(".", "_"))
viper.AutomaticEnv()
```
If we have environment variable `TRANSLATE_SERVICE_PORT=8080`, then it will be available by `viper.GetInt("service.port")`

Precedence does not change **CLI Arguments > environment variables + environment variables from .env > config file**


>How does the binding work with `viper.BindPFlags()`? Exact names? Can we specify the binding function?

In nutshell `viper.BindPFlags()` binds **flag name** to flag e.g.
```go
rootCmd.PersistentFlags().Uint("port", 0, "port to run")
```
in this scenario calling `viper.BindPFlags()` is equivalent to binding manually 
[...]

PR #29, opened 2023-02-27
#25 Added configuration from various sources

VladislavsPerkanuks:
#25


.vscode/extensions.json
```diff
@@ -1,6 +1,5 @@
 {
   "recommendations": [
-    "aleksandra.go-group-imports",
```
VladislavsPerkanuks, line 3: Deleted after discussion with @janishorsts

cmd/translate/main.go
```diff
@@ -122,27 +124,33 @@ func main() {
...
+	} else {
+		fmt.Println("Using config file:", viper.ConfigFileUsed())
+	}
 
-	// If a config file is found, read it in.
-	if err := viper.ReadInConfig(); err == nil {
-		fmt.Fprintln(os.Stderr, "Using config file:", viper.ConfigFileUsed())
+	// For now manually bind CLI argument to viper.
+	err := viper.BindPFlag("service.port", rootCmd.Flags().Lookup("port"))
+	if err != nil {
```
VladislavsPerkanuks, line 153: I am open to suggestions, what would be the best way to bind arguments, from options I found

cmd/translate/main_grpc_integration_test.go
```diff
@@ -14,26 +18,65 @@ import (
...
+	addr             string
+	port             string
 
-	conn, err := grpc.DialContext(
-		ctx,
-		"localhost:8080",
+	client pb.TranslateServiceClient
+)
+
+func TestMain(m *testing.M) {
```
VladislavsPerkanuks, line 29: I once again moved `TestMain` function to gRPC integration test. 
**Reason**: The tests were failing because the server was not ready to accept requests.
Fixed with using `grpc.DialContext` with `grpc.WithBlock` which blocks execution while connection is not established

cmd/translate/main_grpc_integration_test.go
```diff
@@ -14,26 +18,65 @@ import (
...
+
+	wg.Add(1)
+
+	go func() {
+		defer wg.Done()
+		main()
+	}()
+
+	// Wait until viper finishes initialization.
+	for viper.GetUint("service.port") == 0 {
```
VladislavsPerkanuks, line 40: I did not find a better solution to find when viper is finished initializing. Without this waiting `service.port` is 0, e.g. default value.
I tried using the bool flag to indicate that viper values are set in main.go `init()` but couldn't get it to work for some reason

.env.test
```diff
@@ -4,3 +4,7 @@ EARTHLY_INSTALLATION_NAME=translate
 OTEL_SERVICE_NAME=translate
 OTEL_EXPORTER_OTLP_INSECURE=true
 OTEL_RESOURCE_ATTRIBUTES=deployment.environment=test
+
+TRANSLATE_SERVICE_PORT=8080
+TRANSLATE_SERVICE_ADDRESS=localhost
```
janishorsts, line 9: In general, the address is used for <host>:<port>. Use HOST instead.

```suggestion
TRANSLATE_SERVICE_HOST=localhost
```

cmd/translate/main.go
```diff
@@ -122,27 +124,33 @@ func main() {
 func init() {
 	cobra.OnInitialize(initConfig)
 	rootCmd.PersistentFlags().StringVar(&cfgFile, "config", "", "config file (default is ./translate.yaml)")
```
janishorsts, line 126: Set default value

```suggestion
	rootCmd.PersistentFlags().StringVar(&cfgFile, "config", "translate.yaml", "config file (default is ./translate.yaml)")
```

cmd/translate/main.go
```diff
@@ -122,27 +124,33 @@ func main() {
...
+	viper.SetConfigFile(cfgFile)
+
+	viper.SetEnvPrefix("translate")
+	// Replace underscores with dots in environment variable names.
+	viper.SetEnvKeyReplacer(strings.NewReplacer(".", "_"))
+	viper.AutomaticEnv()
+
+	// Try to read config.
+	if err := viper.ReadInConfig(); err != nil {
+		fmt.Fprintln(os.Stderr, err)
```
janishorsts, line 146: panic if we cfgFile is not equal to the default value and it fails to read the file.

cmd/translate/main.go
```diff
@@ -122,27 +124,33 @@ func main() {
...
+		fmt.Println("Using config file:", viper.ConfigFileUsed())
+	}
 
-	// If a config file is found, read it in.
-	if err := viper.ReadInConfig(); err == nil {
-		fmt.Fprintln(os.Stderr, "Using config file:", viper.ConfigFileUsed())
+	// For now manually bind CLI argument to viper.
+	err := viper.BindPFlag("service.port", rootCmd.Flags().Lookup("port"))
+	if err != nil {
+		fmt.Fprintln(os.Stderr, err)
```
janishorsts, line 154: the same, stop execution if an error occurs

cmd/translate/main_grpc_integration_test.go
janishorsts: There are certain values in the integration tests we can determine on the fly. E.g. free port number for the server to listen on.

We run tests on every save. If the previous test execution is still running, a new one will fail - it cannot listen on the same port which is already in use. But by using a random free port for each test execution we can avoid this conflict. Make the test run more resilient on developers' machines.

janishorsts (changes requested):

cmd/translate/main.go
VladislavsPerkanuks: why couldn't be simply?
```go
if err := viper.ReadInConfig(); err != nil {
    log.panic(err)
}
```
In your case what happens, if the default value e.g. `./translate.yaml` can't be read ? (Most likely due to no such file/directory)

.env.test
```diff
@@ -1,6 +1,3 @@
-EARTHLY_ALLOW_PRIVILEGED=true
-EARTHLY_INSTALLATION_NAME=translate
-
```
VladislavsPerkanuks, line 3: Should not belong to testing environment

cmd/translate/main.go
```diff
@@ -121,28 +122,33 @@ func main() {
...
+	viper.SetConfigFile(cfgFile)
+
+	viper.SetEnvPrefix("translate")
+	// Replace underscores with dots in environment variable names.
+	viper.SetEnvKeyReplacer(strings.NewReplacer(".", "_"))
+	viper.AutomaticEnv()
+
+	// Try to read config.
+	if err := viper.ReadInConfig(); err != nil && cfgFile != "translate.yaml" {
+		log.Panic(err)
```
janishorsts, line 141: wrap error

janishorsts (approved):

PR #31, opened 2023-03-03
18-transform-to-and-from-angularlocalize

VladislavsPerkanuks:


pkg/convert/json_ngxtranslate_test.go
```diff
@@ -94,7 +94,8 @@ func Test_FromNgxTranslate(t *testing.T) {
 				return
 			}
 
-			assert.Equal(t, tt.expected, result)
+			assert.Equal(t, tt.expected.Language, result.Language)
+			assert.ElementsMatch(t, tt.expected.Messages, result.Messages)
```
VladislavsPerkanuks, line 98: Fixed random test failing, due to `result.Messages` order could be different than in `tt.expected.Messages`, which cause `assert.Equal()` to fail

pkg/convert/angular_localize_test.go
```diff
@@ -0,0 +1,241 @@
...
+			if !assert.NoError(t, err) {
+				return
+			}
+
+			// Matches zero or more whitespace characters.
+			re := regexp.MustCompile(`\s*`)
+			resultTrimmed := re.ReplaceAllString(string(result), "")
+			wantTrimmed := re.ReplaceAllString(string(tt.want), "")
+
+			assert.Equal(t, resultTrimmed, wantTrimmed)
```
VladislavsPerkanuks, line 238: It is not the best way to assert, but it's the only way I found. Both `result` and `tt.want` are undergoing the same operation e.g. removing all spaces from the string. 
If their trimmed versions match, then their untrimmed versions must also match.

VladislavsPerkanuks: Changed ng to return json format, not relevant anymore

pkg/server/translate/v1/translate_grpc.pb.go
```diff
@@ -1,6 +1,6 @@
 // Code generated by protoc-gen-go-grpc. DO NOT EDIT.
```
VladislavsPerkanuks, line 1: A side effect of upgrading `bufbuild`, integration tests passing

pkg/convert/angular_localize.go
```diff
@@ -0,0 +1,124 @@
...
+
+func (n ngJSON) fromNgJSON() model.Messages {
+	messages := model.Messages{
+		Language: n.Language,
+		Messages: make([]model.Message, 0, len(n.Translations)),
+	}
+
+	for k, v := range n.Translations {
+		msg := model.Message{ID: k, Message: v}
+		messages.Messages = append(messages.Messages, msg)
```
janishorsts, line 33: Personally, I don't bother declaring a new variable unless it adds some clarity and improves readability. Up to you what to do.

```suggestion
		messages.Messages = append(messages.Messages, model.Message{ID: k, Message: v})
```

pkg/convert/angular_localize.go
```diff
@@ -0,0 +1,124 @@
...
+
+	return messages
+}
+
+// FromNG converts serialized data from the ng extract-i18n tool into a model.Messages struct.
+//
+// This function supports two input formats:
+//   - JSON (produced by running "ng extract-i18n --format json")
+//   - XLF (based on the XLIFF 1.2 standard, which is the default format for ng extract-i18n).
+func FromNG(data []byte) (model.Messages, error) {
```
janishorsts, line 80:
```go
// angular_localize.go
func FromAngularLocalize()
func ToAngularLocalize()

// xlf.go (or xfl12.go)
func FromXlf()
func ToXlf()
// or
func FromXlf12()
func ToXlf12()

File: pkg/convert/xliff12_test.go [hunk 6]
```diff
@@ -0,0 +1,175 @@
...
+			}
+
+			if !assert.NoError(t, err) {
+				return
+			}
+
+			// Matches zero or more whitespace characters.
+			re := regexp.MustCompile(`\s*`)
+			resultTrimmed := re.ReplaceAllString(string(result), "")
+			wantTrimmed := re.ReplaceAllString(string(tt.want), "")
```
VladislavsPerkanuks, line 170: It is not the best way to assert, but it's the only way I found. Both result and tt.want are undergoing the same operation which is removing all spaces from the string.
If their trimmed versions match, then their untrimmed versions must also match.

pkg/convert/angular_localize.go
VladislavsPerkanuks: XLF is just a file extension, the standard is **XLIFF 1.2** 
The file extension for XLIFF 1.2 can be `.xliff`, `.xlf` or  even`.xml`

pkg/convert/arb.go
```diff
@@ -63,7 +64,35 @@ func FromArb(data []byte) (model.Messages, error) {
...
+	// ARB file can have optional '@@locale' top level key for source key language.
+	// https://medium.com/@Albert221/how-to-internationalize-your-flutter-app-with-arb-files-today-full-blown-tutorial-476ee65ecaed
+	findLocale := func() (language.Tag, error) {
+		// if '@@locale' is missing then language it is not provided (Undetermined).
+		if _, ok := dst["@@locale"]; !ok {
+			return language.Tag{}, nil
+		}
+
+		// Check if @@locale key's value type is string.
+		langString, ok := dst["@@locale"].(string)
```
janishorsts, line 77:
```suggestion
		locale, ok := dst["@@locale"]
		if !ok {
			return language.Tag{}, nil
		}

		// Check if @@locale key's value type is string.
		langString, ok := locale.(string)
```

pkg/convert/arb_test.go
```diff
@@ -92,6 +94,58 @@ func Test_FromArb(t *testing.T) {
...
+			name: "With locale",
+			data: []byte(`
+      {
+        "@@locale": "en",
+        "title": "Hello World!",
+        "@title": {
+          "description": "Message to greet the World"
+        }
+      }
+			`),
```
janishorsts, line 107: Why do we have 3 tabs on the last line?

janishorsts, line 107: ?

```suggestion
      }`),
```

VladislavsPerkanuks: I think this
image
looks cleaner than this
image

GitHub makes different indentation than vscode

pkg/convert/json_ngxtranslate_test.go
janishorsts: In golang, the key order in the map is indeterministic.

VladislavsPerkanuks: it is not a map, it is `[]model.Message`. 
```go
assert.Equal({msg1, msg2}, {msg2, msg1}) = false

assert.ElementsMatch({msg1, msg2}, {msg2, msg1}) = true
```

pkg/convert/xliff12_test.go
janishorsts: Are you attempting to remove whitespace text nodes between XML elements?

pkg/convert/json_ngxtranslate_test.go
janishorsts: yeah, but under the hood, it uses maps in the flow, doesn't it? :)

pkg/convert/xliff12_test.go
VladislavsPerkanuks: Initially, I wanted to do that, but I could not think of a way to implement that.
Now I am removing all whitespaces from XML document even inside XML elements. 
As we need only to compare one to the other, that cheap hack should suffice.

pkg/convert/json_ngxtranslate_test.go
VladislavsPerkanuks: did not catch that

pkg/convert/xliff12.go
```diff
@@ -0,0 +1,86 @@
...
+			Note:   msg.Description,
+		})
+	}
+
+	data, err := xml.Marshal(&xlf)
+	if err != nil {
+		return nil, fmt.Errorf("marshal xliff12 struct to XLIFF 1.2 formatted XML: %w", err)
+	}
+
+	dataWithHeader := append([]byte(xml.Header), data...) // prepend generic XML header
```
VladislavsPerkanuks, line 83: https://github.com/translate-agent/translate/pull/33#discussion_r1129215996

janishorsts (approved):

Issue #34, opened 2023-03-08
Transform to and from XMB


DarjaLeonova:

janishorsts:
Not widely supported translation standard.

DarjaLeonova:
After investigation:
There is a lot of versions of XMB and which one is the correct we don't know for now. @janishorsts decided to put it on hold

PR #35, opened 2023-03-09
#34 transform to and from XMB

DarjaLeonova:


pkg/convert/xmb.go
```diff
@@ -0,0 +1,71 @@
...
+	ID          string `xml:"id,attr"`
+	Description string `xml:"desc,attr"`
+	Content     string `xml:"content"`
+	Fuzzy       bool   `xml:"fuzzy,attr"`
+}
+
+func FromXMB(b []byte) (model.Messages, error) {
+	var xmbFormat xmb
+
+	err := xml.Unmarshal(b, &xmbFormat)
```
klavinsjanis, line 27: alternative:
`	if err := xml.Unmarshal(b, &xmbFormat); err != nil {
		return model.Messages{}, fmt.Errorf("unmarshaling xmb to model.Messages: %w", err)
	}`

pkg/convert/xmb.go
```diff
@@ -0,0 +1,71 @@
...
+	if err != nil {
+		return model.Messages{}, fmt.Errorf("unmarshaling xmb to model.Messages: %w", err)
+	}
+
+	messages := model.Messages{Messages: make([]model.Message, 0, len(xmbFormat.Messages))}
```
klavinsjanis, line 32: Why use different implementations, reserving slice capacity in one and length in other.
`	xmbFormat := xmb{
		Messages: make([]msg, len(m.Messages)),
		Language: m.Language,
	}`

pkg/convert/xmb.go
```diff
@@ -0,0 +1,71 @@
...
+	Language language.Tag `xml:"srcLang,attr"`
+	Messages []msg        `xml:"message"`
+}
+
+type msg struct {
+	ID          string `xml:"id,attr"`
+	Description string `xml:"desc,attr"`
+	Content     string `xml:"content"`
+	Fuzzy       bool   `xml:"fuzzy,attr"`
+}
```
VladislavsPerkanuks, line 22: 1) There is no `Language` in any form (element or attr) in XMB formatted XML, from my research
2) The resources are very sparse, but I found [example](https://localizify.com/formats/angular-xmb) of how XMB looks like, and it is kinda different from what you have now

pkg/convert/xmb.go
```diff
@@ -0,0 +1,71 @@
...
+
+func FromXMB(b []byte) (model.Messages, error) {
+	var xmbFormat xmb
+
+	err := xml.Unmarshal(b, &xmbFormat)
+	if err != nil {
+		return model.Messages{}, fmt.Errorf("unmarshaling xmb to model.Messages: %w", err)
```
VladislavsPerkanuks, line 29: You are not unmarshaling to `model.Messages`, but to `xmb`

pkg/convert/xmb.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+type msg struct {
+	ID          string `xml:"id,attr"`
+	Description string `xml:"desc,attr"`
+	Content     string `xml:"content"`
+	Fuzzy       bool   `xml:"fuzzy,attr"`
+}
+
+func FromXMB(b []byte) (model.Messages, error) {
```
VladislavsPerkanuks, line 24:
```suggestion
func FromXmb(b []byte) (model.Messages, error) {
```
Consistency with other From's

pkg/convert/xmb.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+
+	return model.Messages{
+		Language: xmbFormat.Language,
+		Messages: messages.Messages,
+	}, nil
+}
+
+func ToXMB(m model.Messages) ([]byte, error) {
+	xmbFormat := xmb{
```
klavinsjanis, line 49: xmbFormat -> xmb?

pkg/convert/xmb_test.go
```diff
@@ -0,0 +1,176 @@
...
+
+	if !assert.NoError(t, err) {
+		return
+	}
+
+	r := regexp.MustCompile(`>(\s*)<`)
+	actualTrimmed := r.ReplaceAllString(string(actual), "><")
+	expectedTrimmed := r.ReplaceAllString(string(expected), "><")
+
+	assert.Equal(t, expectedTrimmed, actualTrimmed)
```
VladislavsPerkanuks, line 175:
```suggestion
	assertEqualXml(t, tt.expected, actual)
```
Merge main if not done already

pkg/convert/xmb.go
VladislavsPerkanuks: Here is another [example](https://github.com/angular/angular/issues/12441#issuecomment-261026476)

DarjaLeonova: After investigation: 
There is a lot of versions of XMB and which one is the correct we don't know for now. @janishorsts decided to put it on hold

VladislavsPerkanuks:
removing clutter from PR's :)
When it will be time, you should create new PR, as things probably gonna change

PR #37, opened 2023-03-09
#36 Initial repo package

VladislavsPerkanuks:
#36

### DIFF
--- a/pkg/translate/translate.go
+++ b/pkg/translate/translate.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 
+	"go.expect.digital/translate/pkg/repo"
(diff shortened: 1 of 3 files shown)


pkg/translate/translate.go
```diff
@@ -13,6 +14,11 @@ import (
 
 type TranslateServiceServer struct {
 	pb.UnimplementedTranslateServiceServer
+	repo repo.Repo
+}
```
VladislavsPerkanuks, line 18: As far as I understood repo field, holds a copy of repo, not the repo instance itself,
Maybe it should hold a pointer to `repo.Repo` ?
```go
type TranslateServiceServer struct {
	pb.UnimplementedTranslateServiceServer
	repo *repo.Repo
}
```

PR #41, opened 2023-03-13
40 - add MySQL db

janishorsts:


janishorsts:
Not required anymore. Already cherry-picked into other branches.

PR #42, opened 2023-03-13
39-add-service-entity-endpoints

VladislavsPerkanuks:


Earthfile
```diff
@@ -37,6 +37,7 @@ proto:
   ' gen/proto/go/translate/v1/translate.pb.gw.go
 
   RUN rm gen/proto/go/translate/v1/translate.pb.gw.go.bak
+  #TODO Save to more apropriate dir e.g. pkg/pb/translate/v1 or pkg/gen/translate/v1
```
VladislavsPerkanuks, line 40: In the future want to refractor the structure a bit for more clarity:
1. Change `api/translate/translate/v1/translate.proto` to `proto/translate/v1/translate.proto`
2. Change `pkg/server/translate/v1...` to `pkg/pb/translate/v1...`

Too many changes for this PR

pkg/translate/file.go
```diff
@@ -0,0 +1,118 @@
...
+	"golang.org/x/text/language"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+	"google.golang.org/protobuf/types/known/emptypb"
+)
+
+type (
+	uploadTranslationFileRequest   pb.UploadTranslationFileRequest
+	downloadTranslationFileRequest pb.DownloadTranslationFileRequest
+)
```
VladislavsPerkanuks, line 18: I chose to define new types so that we can create methods for them. 
This approach eliminates the need to create a new function, such as `parseUploadParams` or `parseDownloadParams`, for parsing the parameters for each endpoint. 
Instead, we can parse the parameters using the `.parseParams()` method, without specifying which request we are parsing for explicitly.

api/translate/translate/v1/translate.proto
```diff
@@ -40,7 +47,32 @@ message DownloadTranslationFileResponse {
   bytes data = 1;
 }
 
+// -----------------Service requests/responses-----------------------
+
+message LoadServiceRequest {
```
janishorsts, line 52: Please prefix with `Get`, `Create`, `List`, `Update` and `Delete` instead.

pkg/translate/service.go
```diff
@@ -0,0 +1,160 @@
...
+type deleteServiceParams struct {
+	id uuid.UUID
+}
+
+func (d *deleteServiceRequest) parseParams() (deleteServiceParams, error) {
+	if d == nil {
+		return deleteServiceParams{}, errors.New("request is nil")
+	}
+
+	uuid, err := uuid.Parse(d.Id)
```
janishorsts, line 131: Try to avoid shadowing names. Simply `id`.

```suggestion
	id, err := uuid.Parse(d.Id)
```

janishorsts (approved):

PR #44, opened 2023-03-16
40-add-service-repo

VladislavsPerkanuks:
Actual implementation to endpoints, e2e tests and earthly changes in different pull request/requests to avoid too many changes in one PR


pkg/convert/pot_test.go
```diff
@@ -334,7 +334,6 @@ msgstr "Au revoir!"
 				return
 			}
 
-			fmt.Printf("res: %s", result)
```
VladislavsPerkanuks, line 337: Redundant print, additional undesirable noise for tests

pkg/repo/mysql/mysql.go
```diff
@@ -0,0 +1,34 @@
...
+type Conf struct {
+	Addr     string
+	User     string
+	Password string
+	Database string
+}
+
+func DefaultConf() (*Conf, error) {
+	return new(Conf), nil
+}
```
VladislavsPerkanuks, line 20: Maybe DefaultConf should return
```go
	return &Conf{
		Addr:     "localhost:3306",
		User:     "root",
		Database: "translate",
	}
```
And in main we could unmarshal viper to Conf struct. need to think about it. 
Maybe some suggestions are available

db/mysql/migrate/000001_initial_schema.up.sql
```diff
@@ -0,0 +1 @@
+CREATE TABLE IF NOT EXISTS Service(ID varchar(36) PRIMARY KEY, Name VARCHAR(255));
```
klavinsjanis, line 1: MySQL doesn't have UUID type?

VladislavsPerkanuks: nope

go.sum
```diff
@@ -1,3 +1,5 @@
+bazil.org/fuse v0.0.0-20160811212531-371fbbdaa898/go.mod h1:Xbm+BRKSBEpa4q4hTSxohYNQpsxXPbPry4JJWOB3LB8=
```
klavinsjanis, line 1: 1335 new lines after go mod tidy?

VladislavsPerkanuks: yup something off. Forgot to run tidy after deleting some packages for local testing or something.

pkg/repo/mysql/mysql.go
```diff
@@ -0,0 +1,34 @@
...
+import (
+	"context"
+	"database/sql"
+	"fmt"
+
+	_ "github.com/go-sql-driver/mysql"
+)
+
+type Conf struct {
+	Addr     string
```
klavinsjanis, line 12: maybe best to separate address into two fields 'host' and 'port' for clarify.

pkg/repo/mysql/mysql.go
```diff
@@ -0,0 +1,34 @@
...
+	Password string
+	Database string
+}
+
+func DefaultConf() (*Conf, error) {
+	return new(Conf), nil
+}
+
+func NewDB(ctx context.Context, conf *Conf) (*sql.DB, error) {
+	db, err := sql.Open("mysql", fmt.Sprintf("%s:%s@tcp(%s)/%s", conf.User, conf.Password, conf.Addr, conf.Database))
```
klavinsjanis, line 23: There should be OTEL wrapper for sql.Open.

pkg/repo/mysql/repo.go
```diff
@@ -0,0 +1,58 @@
+package mysql
+
+import (
+	"context"
+	"database/sql"
+	"fmt"
+)
+
+type Repo struct {
```
klavinsjanis, line 9: Why Repo struct, are you planning to create methods?

pkg/repo/mysql/repo.go
```diff
@@ -0,0 +1,58 @@
+package mysql
```
klavinsjanis, line 1: package mysql? That might conflict with imported mysql package, could be worth keeping the package name vague like for example - package db

pkg/repo/mysql/services.go
```diff
@@ -0,0 +1,92 @@
...
+	case errors.Is(err, sql.ErrNoRows):
+		return nil, repo.ErrNotFound
+	case err != nil:
+		return nil, fmt.Errorf("db: select service: %w", err)
+	}
+
+	return &service, nil
+}
+
+func (r *Repo) LoadServices(ctx context.Context) ([]model.Service, error) {
```
klavinsjanis, line 44: Doesn't looks like a LoadServices function, more like GetServices.

pkg/repo/mysql/services_integration_test.go
```diff
@@ -0,0 +1,213 @@
...
+	viper.SetEnvPrefix("translate_mysql")
+	viper.AutomaticEnv()
+
+	mysqlConf := Conf{
+		Addr:     viper.GetString("address"),
+		User:     viper.GetString("user"),
+		Database: viper.GetString("database"),
+	}
+
+	db, err := NewDB(context.Background(), &mysqlConf)
```
klavinsjanis, line 28: why no context propagation?

VladislavsPerkanuks: From where ? This is top level ctx

.env.test
```diff
@@ -1,3 +1,7 @@
 OTEL_SERVICE_NAME=translate
 OTEL_EXPORTER_OTLP_INSECURE=true
 OTEL_RESOURCE_ATTRIBUTES=deployment.environment=test
+
+TRANSLATE_MYSQL_ADDRESS=localhost:3306
```
klavinsjanis, line 5: Why use TRANSLATE_ prefix? are there any other services?

pkg/repo/mysql/repo.go
VladislavsPerkanuks: shouldn't conflict, as they never be imported both in the same file

.env.test
VladislavsPerkanuks: Future proof

pkg/repo/mysql/mysql.go
VladislavsPerkanuks: I guess, it will be better

pkg/repo/mysql/services.go
VladislavsPerkanuks: Can't say anything about that, but I also like idea of `Create, Delete, Get, List` more.
Those names that are now came from Horsts, personally don't know which are better

pkg/repo/mysql/services_integration_test.go
klavinsjanis: might be worth to pass the context through the functions in this function scope, just looks :)

VladislavsPerkanuks: Sure

pkg/repo/mysql/services_integration_test.go
```diff
@@ -0,0 +1,213 @@
...
+  (
+    '11111111-1111-1111-1111-111111111111',
+    'Service 2'
+  ),
+  (
+    '22222222-2222-2222-2222-222222222222',
+    'Service 3'
+  );`
+
+	_, err = db.ExecContext(context.Background(), insertQuery)
```
klavinsjanis, line 47: db.ExecContext(ctx, insertQuery)

pkg/repo/mysql/services_integration_test.go
VladislavsPerkanuks: I suppose they should have tracing enabled

pkg/repo/mysql/mysql.go
```diff
@@ -0,0 +1,34 @@
...
+	User     string
+	Password string
+	Database string
+}
+
+func DefaultConf() (*Conf, error) {
+	return new(Conf), nil
+}
+
+func NewDB(ctx context.Context, conf *Conf) (*sql.DB, error) {
```
klavinsjanis, line 22: shouldn't you establish db connection in main.go or just using for testing now?

pkg/repo/mysql/services_integration_test.go
```diff
@@ -0,0 +1,213 @@
...
+
+	"github.com/brianvoe/gofakeit/v6"
+	"github.com/google/uuid"
+	"github.com/spf13/viper"
+	"github.com/stretchr/testify/assert"
+	"go.expect.digital/translate/pkg/model"
+	"go.expect.digital/translate/pkg/repo"
+)
+
+var mysqlRepo *Repo
```
klavinsjanis, line 16: alternative:
var repo *Repo

pkg/repo/mysql/mysql.go
```diff
@@ -0,0 +1,34 @@
+package mysql
```
klavinsjanis, line 1: To avoid nesting:
pkg/repo/repo.go

pkg/repo/mysql/services.go
klavinsjanis: Maybe LoadServices was the handler name not for db func?

VladislavsPerkanuks: https://github.com/translate-agent/translate/issues/40#issue-1620982418

pkg/repo/mysql/mysql.go
VladislavsPerkanuks: Found 3 packages, this is the one I liked the most https://github.com/XSAM/otelsql
Maybe in the future we will need to do manual instrumenting, because all those packages are not super configurable

db/mysql/migrate/000001_initial_schema.up.sql
janishorsts, line 1: We use snake case naming for identifiers in databases.

In MySQL, the table names rely on the underlying system https://dev.mysql.com/doc/refman/8.0/en/identifier-case-sensitivity.html

```suggestion
CREATE TABLE service(id char(36) PRIMARY KEY, name varchar(255));
```

janishorsts, line 1: Let's add SQL linter. I suggest https://github.com/sqlfluff/sqlfluff but if you'd like an alternative, please let me know.

db/mysql/migrate/000001_initial_schema.down.sql
```diff
@@ -0,0 +1 @@
+DROP TABLE IF EXISTS Service;
```
janishorsts, line 1:
```suggestion
DROP TABLE service;
```

pkg/repo/mysql/mysql.go
janishorsts: Think SSOT.

db/mysql/migrate/000001_initial_schema.up.sql
VladislavsPerkanuks: Will look into that

pkg/repo/mysql/services.go
```diff
@@ -0,0 +1,92 @@
...
+	"fmt"
+
+	"github.com/google/uuid"
+	"go.expect.digital/translate/pkg/model"
+	"go.expect.digital/translate/pkg/repo"
+)
+
+func (r *Repo) SaveService(ctx context.Context, service *model.Service) error {
+	query := `INSERT INTO Service (ID, Name) VALUES (?, ?)
+	ON DUPLICATE KEY UPDATE Name = VALUES(Name)`
```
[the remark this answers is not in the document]
janishorsts:
```sql
-- the SQL statement displayed in traces UI (jaeger, kibana etc)
-- formatted by you
INSERT INTO Service (ID, Name) VALUES (?, ?)
	ON DUPLICATE KEY UPDATE Name = VALUES(Name)
-- vs mine
INSERT INTO service (id, name) VALUES (?, ?) ON DUPLICATE KEY UPDATE name = VALUES(name)
```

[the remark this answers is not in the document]
VladislavsPerkanuks: Noted

pkg/repo/mysql/services.go
```diff
@@ -0,0 +1,92 @@
...
+	err := row.Scan(&service.ID, &service.Name)
+
+	switch {
+	case errors.Is(err, sql.ErrNoRows):
+		return nil, repo.ErrNotFound
+	case err != nil:
+		return nil, fmt.Errorf("db: select service: %w", err)
+	}
+
+	return &service, nil
```
janishorsts, line 41: This is how I do it, you don't have to follow it. `default` case is at the top to ease the reading of the happy path, bool switch provides a clear view of different outcomes.

```suggestion
	err := row.Scan(&service.ID, &service.Name)
	switch {
	default:
		return &service, nil
	case errors.Is(err, sql.ErrNoRows):
		return nil, repo.ErrNotFound
	case err != nil:
		return nil, fmt.Errorf("db: select service: %w", err)
	}	
```

pkg/repo/mysql/services.go
janishorsts: In repo, `Load` and `Save` are alternatives to `Get` and `Create/Update`. They convey the same meaning. The aim is to avoid defining two repo methods to create or update an entity. 

In the past, I always used `Create` and `Update` but I feel we can get away with `Save` to simplify the repo interface.

pkg/repo/mysql/services.go
```diff
@@ -0,0 +1,92 @@
...
+	defer rows.Close()
+
+	var services []model.Service
+
+	for rows.Next() {
+		var service model.Service
+
+		err = rows.Scan(&service.ID, &service.Name)
+		if err != nil {
+			return nil, fmt.Errorf("scan row: %w", err)
```
janishorsts, line 60:
```suggestion
			return nil, fmt.Errorf("scan service: %w", err)
```

pkg/repo/mysql/services.go
```diff
@@ -0,0 +1,92 @@
...
+		}
+
+		services = append(services, service)
+	}
+
+	err = rows.Err()
+	if err != nil {
+		return nil, fmt.Errorf("iterate rows: %w", err)
```
janishorsts, line 68:
```suggestion
		return nil, fmt.Errorf("scan services: %w", err)
```

pkg/repo/mysql/services.go
janishorsts, line 1: Please use the prefix `db: ` in all error messages to be consistent.

pkg/repo/mysql/services.go
VladislavsPerkanuks: okay, I am just used to `default` always being the last option in switch cases, not sure which option  I like best.

pkg/repo/mysql/services.go
```diff
@@ -0,0 +1,92 @@
...
+	query := `DELETE FROM Service WHERE ID = ?`
+
+	result, err := r.db.ExecContext(ctx, query, serviceID)
+	if err != nil {
+		return fmt.Errorf("db: delete service: %w", err)
+	}
+
+	count, err := result.RowsAffected()
+	if err != nil {
+		return fmt.Errorf("retrieve affected rows: %w", err)
```
janishorsts, line 84: Alternatively, to provide better context for the error message.

```suggestion
		return fmt.Errorf("db: is service deleted: %w", err)
```

VladislavsPerkanuks: error message in question form?

pkg/repo/mysql/services_integration_test.go
```diff
@@ -0,0 +1,213 @@
...
+		User:     viper.GetString("user"),
+		Database: viper.GetString("database"),
+	}
+
+	db, err := NewDB(context.Background(), &mysqlConf)
+	if err != nil {
+		log.Panicf("create new db: %v", err)
+	}
+
+	insertQuery := `REPLACE INTO Service (ID, Name)
```
janishorsts, line 43: You must isolate all tests, no shared data etc. Use helper functions to populate entities as and when needed in tests.

pkg/repo/mysql/services_integration_test.go
```diff
@@ -0,0 +1,213 @@
...
+	ctx := context.Background()
+
+	tests := []struct {
+		input *model.Service
+		name  string
+	}{
+		{
+			name: "Save Service",
+			input: &model.Service{
+				ID:   uuid.MustParse(gofakeit.UUID()),
```
janishorsts, line 90:
```suggestion
				ID:   uuid.New(),
```

pkg/repo/mysql/services.go
janishorsts: I like it. It brings to attention what is the default outcome, followed by alternatives and unhappy paths. All outcomes are laid out in parallel - only one path is possible, with less code.

db/mysql/migrate/000001_initial_schema.up.sql
VladislavsPerkanuks: Checked other linters, sqlfluff seems superior in all aspects:

- multiple dialects
- configurable
- vscode extension
- most stars compared to other linters, and actively mainained

pkg/repo/mysql/services_integration_test.go
VladislavsPerkanuks: Name conflict with package repo

pkg/repo/mysql/mysql.go
```diff
@@ -0,0 +1,66 @@
...
+
+import (
+	"context"
+	"database/sql"
+	"fmt"
+
+	"github.com/XSAM/otelsql"
+	_ "github.com/go-sql-driver/mysql"
+	"github.com/spf13/viper"
+	semconv "go.opentelemetry.io/otel/semconv/v1.4.0"
```
klavinsjanis, line 11: v1.4.0 isn't there are newer version?

pkg/repo/mysql/services.go
```diff
@@ -0,0 +1,88 @@
...
+	query := `DELETE FROM service WHERE id = ?`
+
+	result, err := r.db.ExecContext(ctx, query, serviceID)
+	if err != nil {
+		return fmt.Errorf("db: delete service: %w", err)
+	}
+
+	count, err := result.RowsAffected()
+	if err != nil {
+		return fmt.Errorf("db: is service deleted: %w", err)
```
klavinsjanis, line 80: is service deleted? a bit confusing error

pkg/repo/mysql/mysql.go
VladislavsPerkanuks: vscode automatic import, is silly sometimes, but you are right

pkg/repo/mysql/services_integration_test.go
```diff
@@ -0,0 +1,222 @@
...
+
+var mysqlRepo *Repo
+
+func TestMain(m *testing.M) {
+	ctx := context.Background()
+
+	viper.SetEnvPrefix("translate_mysql")
+	viper.AutomaticEnv()
+
+	tp, err := tracer.TracerProvider()
```
klavinsjanis, line 27: otel.SetTracerProvider(tp) to set tracer provider globally?

pkg/repo/mysql/services_integration_test.go
```diff
@@ -0,0 +1,222 @@
...
+
+func createTestService() *model.Service {
+	return &model.Service{
+		Name: gofakeit.FirstName(),
+		ID:   uuid.New(),
+	}
+}
+
+// Inserts service to db.
+func insertTestService(t *testing.T, service *model.Service) error {
```
klavinsjanis, line 63: function comments should start with function name and then descripiton.

pkg/repo/mysql/services.go
klavinsjanis: how about `db: delete service result`

pkg/repo/mysql/services_integration_test.go
VladislavsPerkanuks: it is already :)

pkg/repo/mysql/services.go
VladislavsPerkanuks: yeah, I like to better

pkg/repo/mysql/services_integration_test.go
```diff
@@ -0,0 +1,222 @@
...
+		Name: gofakeit.FirstName(),
+		ID:   uuid.New(),
+	}
+}
+
+// insertTestService inserts service to DB.
+func insertTestService(t *testing.T, service *model.Service) error {
+	t.Helper()
+
+	_, err := mysqlRepo.db.Exec(`INSERT INTO service (id, name) VALUES (?, ?)`, service.ID, service.Name)
```
klavinsjanis, line 66: is there ExecContext available?

klavinsjanis (approved):

pkg/repo/mysql/services_integration_test.go
```diff
@@ -0,0 +1,224 @@
...
+
+	_, err := mysqlRepo.db.ExecContext(ctx, `INSERT INTO service (id, name) VALUES (?, ?)`, service.ID, service.Name)
+	if err != nil {
+		return fmt.Errorf("insert test service: %w", err)
+	}
+
+	return nil
+}
+
+func Test_MysqlSaveService(t *testing.T) {
```
janishorsts, line 74: No need to have Mysql in the test function name. We know it is `mysql` package. Please update all test function names.

```suggestion
func Test_SaveService(t *testing.T) {
```

pkg/repo/mysql/services.go
```diff
@@ -0,0 +1,88 @@
...
+	"errors"
+	"fmt"
+
+	"github.com/google/uuid"
+	"go.expect.digital/translate/pkg/model"
+	"go.expect.digital/translate/pkg/repo"
+)
+
+func (r *Repo) SaveService(ctx context.Context, service *model.Service) error {
+	query := `INSERT INTO service (id, name) VALUES (?, ?) ON DUPLICATE KEY UPDATE name = VALUES (name)`
```
janishorsts, line 15: We need to ensure we set id if not set.

```go
if service.ID == uuid.Nil {
  service.ID = uuid.New()
}
```

pkg/repo/mysql/services.go
```diff
@@ -0,0 +1,88 @@
...
+	count, err := result.RowsAffected()
+	if err != nil {
+		return fmt.Errorf("db: delete service result: %w", err)
+	}
+
+	if count == 0 {
+		return repo.ErrNotFound
+	}
+
+	return nil
```
janishorsts, line 87: Bool switch takes less space and defines all outcomes.

```suggestion
	switch {
	default:
		return nil
	case err != nil:
		return fmt.Errorf("db: delete service result: %w", err)
	case count == 0:
		return repo.ErrNotFound
	}
```

pkg/repo/mysql/services_integration_test.go
janishorsts: Alternative name `repository` to consider instead of `mysqlRepo`.

pkg/repo/mysql/services_integration_test.go
```diff
@@ -0,0 +1,224 @@
...
+	"go.expect.digital/translate/pkg/repo"
+	"go.expect.digital/translate/pkg/tracer"
+)
+
+var mysqlRepo *Repo
+
+func TestMain(m *testing.M) {
+	ctx := context.Background()
+
+	viper.SetEnvPrefix("translate_mysql")
```
janishorsts, line 24: This is a global viper setting. It won't work when we have `viper.SetEnvPrefix("translate")` in the CLI.

pkg/repo/mysql/services_integration_test.go
```diff
@@ -0,0 +1,224 @@
...
+
+func createTestService() *model.Service {
+	return &model.Service{
+		Name: gofakeit.FirstName(),
+		ID:   uuid.New(),
+	}
+}
+
+// insertTestService inserts service to DB.
+func insertTestService(t *testing.T, ctx context.Context, service *model.Service) error {
```
janishorsts, line 63: Doesn't `repository.SaveService()` do the same?

VladislavsPerkanuks: Yes, but I thought that is not a good practice to use one repo function in the other.
If we use `repo.SaveService`, in all other tests "repo.LoadService" etc to insert data before testing actual functionality., then if SaveService does not work as intended then all tests are failing. So that makes them dependent on each other?

pkg/repo/mysql/services_integration_test.go
VladislavsPerkanuks: Despite the name `TestMain()`, this function does not call `main.go main()` (The name must be `TestMain` to pass argument `*testing.M`). I know it is counter-intuitive, but that is one way to set up an integration test for repo: In this case, setting up the tracer and MySQL repo.

So shortly, for service integration tests, there is a separate viper instance, which does not have anything in common with the CLI viper instance, because they will never be called in the same runtime

janishorsts: We should take into account it is the same runtime.

pkg/repo/mysql/services_integration_test.go
janishorsts: As we climb higher and higher in the test pyramid, it becomes harder and harder to test. Keep it simple. If we have an implementation that does the job, we use it - we trust it, and it is tested as well. There is no point in duplicating implementation, now you have to worry about two implementations instead of one, and the effort of maintaining it grows exponentially. The system becomes more fragile.

pkg/repo/mysql/services_integration_test.go
VladislavsPerkanuks: Could you elaborate a little more about that ?
What would be the best course of action?

pkg/repo/mysql/services_integration_test.go
```diff
@@ -0,0 +1,218 @@
...
+}
+
+func Test_SaveService(t *testing.T) {
+	t.Parallel()
+
+	service := randService()
+	err := repository.SaveService(context.Background(), service)
+
+	assert.NoError(t, err)
+}
```
VladislavsPerkanuks, line 64: Is it enough to assert that there is no error after calling `repository.SaveService()` ?
Or should we expand these tests with additional, 
```go
actual, err := repository.LoadService(ctx, service.ID)
...
assert.Equal(t, "loaded from repo", "created randService()")
```

I like It because then we can be 100% sure that everything is working as expected, especially this is for `UpdateService Test`. But I don't like it because we are creating spaghetti of repo methods, in almost all tests for example:
`Test_SaveService = SaveService + LoadService`

VladislavsPerkanuks: I have come to the conclusion that I should assert that the values obtained from the repository match the expected values.

pkg/repo/mysql/services.go
```diff
@@ -0,0 +1,90 @@
...
+	"errors"
+	"fmt"
+
+	"github.com/google/uuid"
+	"go.expect.digital/translate/pkg/model"
+	"go.expect.digital/translate/pkg/repo"
+)
+
+func (r *Repo) SaveService(ctx context.Context, service *model.Service) error {
+	query := `INSERT INTO service (id, name) VALUES (UUID_TO_BIN(?), ?) ON DUPLICATE KEY UPDATE name = VALUES (name)`
```
janishorsts, line 15: I'll try to use `sql.Scanner()` and `driver.Value()` in a separate PR to simplify the SQL statements - to avoid using UUID_TO_BIN and BIN_TO_UUID.

pkg/repo/mysql/services.go
```diff
@@ -0,0 +1,90 @@
...
+
+	if service.ID == uuid.Nil {
+		service.ID = uuid.New()
+	}
+
+	_, err := r.db.ExecContext(ctx, query, service.ID, service.Name)
+	if err != nil {
+		return fmt.Errorf("db: insert service: %w", err)
```
janishorsts, line 23: use `repo` everywhere in the error messages instead of `db`

```suggestion
		return fmt.Errorf("repo: insert service: %w", err)
```

pkg/repo/mysql/services.go
janishorsts: Actually, google uuid already supports sql.Scanner -https://github.com/google/uuid/blob/master/sql.go#L15. No need to convert BIN_TO_UUID when selecting data.

However, sql.Valuer() returns string represantion which requires UUID_TO_BIN https://github.com/google/uuid/blob/master/sql.go#L57

pkg/repo/mysql/services_integration_test.go
```diff
@@ -0,0 +1,254 @@
...
+	}
+
+	repository, err = NewRepo(WithDefaultDB(ctx))
+	if err != nil {
+		log.Panicf("create new repo: %v", err)
+	}
+
+	code := m.Run()
+
+	repository.db.Close()
```
janishorsts, line 41:
```suggestion
	defer repository.db.Close()
	
	code := m.Run()
```

pkg/repo/mysql/services_integration_test.go
```diff
@@ -0,0 +1,254 @@
...
+	actual, err := repository.LoadServices(ctx)
+	if !assert.NoError(t, err) {
+		return
+	}
+
+	for _, expected := range expectedServices {
+		if !assert.Contains(t, actual, expected) {
+			return
+		}
+	}
```
janishorsts, line 203:
```suggestion
	assert.ElementsMatch(t, expected, actual)
```

pkg/repo/mysql/services_integration_test.go
VladislavsPerkanuks: The `defer` keyword does not work with `os.Exit()` because it causes the program to end abruptly.

pkg/repo/mysql/services_integration_test.go
VladislavsPerkanuks: `ElementsMatch` asserts that `list1` contains the same elements as `list2`. So this assertion needs a length of lists to be equal. In this case, the `actual` variable returned by `ListServices()` holds all the services inserted into the database. Since their sizes are not equal (3 vs. n, where n is the number of records in the database), this assertion will not work.

janishorsts (approved):

Issue #30, opened 2023-03-01
Plural translation


janishorsts:
Initially, we require only to be able to parse POT plural, convert it to MessageFormat 2 struct and use `fmt.Stringer` to get string value which we can set in our Message struct.

- add `fmt.Stringer` to MessageFormat 2 struct
- convert POT AST plural to MessageFormat 2 struct

https://github.com/unicode-org/message-format-wg
https://www.gnu.org/software/gettext/manual/html_node/Plural-forms.html

```pot
msgid_plural "foobar"
msgstr[0] "You have {$count} notifications."
msgstr[1] "You have one notification."
msgstr[2..] "{You have {$count} notifications."
```
I'm not sure if `msgstr[2..] "{You have {$count} notifications."` is the correct value in the POT.

```go
message := model.Message{
  ID: "foobar",
  Message: `match {$count :number}
when 1 {You have one notification.}
when * {You have {$count} notifications.}`,
}

PR #45, opened 2023-03-22
#30 plural translation

DarjaLeonova:
I have currently added Plural support for PO only. Now, other formats don't have Plural support. As agreed with Janis, I will investigate other formats and how they support Plural forms, after that we will add Plural support for other formats


pkg/convert/pot.go
```diff
@@ -73,33 +49,21 @@ func FromPot(b []byte) (model.Messages, error) {
...
 
+	for _, node := range po.Messages {
+		var message model.Message
+		if po.Header.PluralForms.NPlurals > 1 { // Handle messages with plural forms
+			message.ID = node.MsgId
+			message.PluralID = node.MsgIdPlural
+			message.Message = convertToFormatMessage(node.MsgStr)
+		} else {
+			message.ID = node.MsgId
 			message.Message = node.MsgStr[0]
```
VladislavsPerkanuks, line 61: You can combine all same fields for struct,
```suggestion
		message := model.Message{ID:node.MsgId}
		if po.Header.PluralForms.NPlurals > 1 { // Handle messages with plural forms
			message.PluralID = node.MsgIdPlural
			message.Message = convertToFormatMessage(node.MsgStr)
		} else {
			message.Message = node.MsgStr[0]
```

pkg/convert/pot.go
```diff
@@ -110,47 +74,209 @@ func FromPot(b []byte) (model.Messages, error) {
...
-		return fmt.Errorf("marshal %s: %w", tag, err)
+	switch tag { //nolint:exhaustive
+	case MsgStrPlural:
+		if err := writePluralForms(b, tag, str); err != nil {
+			return fmt.Errorf("write plural forms: %w", err)
+		}
+	default:
+		if err := writeDefault(b, tag, str); err != nil {
+			return fmt.Errorf("write default: %w", err)
+		}
```
VladislavsPerkanuks, line 85: how about:
```suggestion
	var write func(*bytes.Buffer, poTag, string) error

	switch tag { //nolint:exhaustive
	case MsgStrPlural:
		write = writePluralForms
	default:
		write = writeDefault
	}
	if err := write(b, tag, str); err != nil {
		return fmt.Errorf("write: %w", err)
	}
```

pkg/convert/pot.go
```diff
@@ -110,47 +74,209 @@ func FromPot(b []byte) (model.Messages, error) {
...
+	if err != nil {
+		return fmt.Errorf("get po tag lines: %w", err)
 	}
 
 	if len(lines) == 1 {
-		_, err = fmt.Fprintf(b, "%s \"%s\"\n", tag, lines[0])
-		if err != nil {
+		if _, err = fmt.Fprintf(b, "%s \"%s\"\n", tag, lines[0]); err != nil {
 			return fmt.Errorf("write %s: %w", tag, err)
 		}
```
VladislavsPerkanuks, line 100: Maybe use switch here to avoid nesting ?

```go
switch len(lines){
case 1:
...
default:
...}
```

pkg/convert/pot.go
```diff
@@ -110,47 +74,209 @@ func FromPot(b []byte) (model.Messages, error) {
...
+		return fmt.Errorf("extract plural forms: %w", err)
+	}
+
+	for i := range pluralForms {
+		err, pluralLines := getPoTagLines(pluralForms[i], tag)
+		if err != nil {
+			return fmt.Errorf("get po tag lines: %w", err)
+		}
+
+		if len(pluralLines) == 1 {
```
VladislavsPerkanuks, line 128: also think that here swithc would look more elegant that nested if
```go
switch len(pluralLines){
case 1:
...}
```

pkg/convert/pot.go
```diff
@@ -110,47 +74,209 @@ func FromPot(b []byte) (model.Messages, error) {
...
+			}
+		}
+	}
+
+	return nil
+}
+
+func writeMultiline(b *bytes.Buffer, tag poTag, lines []string) error {
 	for _, line := range lines {
 		if strings.HasSuffix(line, "\\n") {
```
VladislavsPerkanuks, line 147: maybe also `switch  strings.HasSuffix(line, "\\n")`?

pkg/convert/pot.go
```diff
@@ -110,47 +74,209 @@ func FromPot(b []byte) (model.Messages, error) {
...
+			if _, err := fmt.Fprint(b, "\""+line+"\\n\""+"\n"); err != nil {
 				return fmt.Errorf("write %s: %w", tag, err)
 			}
 		}
 	}
 
 	return nil
 }
+
+func getPoTagLines(str string, tag poTag) (error, []string) {
```
VladislavsPerkanuks, line 161: error is last in function arguments
```suggestion
func getPoTagLines(str string, tag poTag) ( []string, error) {
```

pkg/convert/pot.go
```diff
@@ -110,47 +74,209 @@ func FromPot(b []byte) (model.Messages, error) {
...
+		return fmt.Errorf("marshal %s: %w", tag, err), nil
+	}
+
+	encodedStr = encodedStr[1 : len(encodedStr)-1] // trim quotes
+	lines := strings.Split(string(encodedStr), "\\n")
+
+	// Remove the empty string element. The line is empty when splitting by "\\n" and "\n" is the last character in str.
+	if lines[len(lines)-1] == "" {
+		lines = lines[:len(lines)-1]
+		lines[len(lines)-1] = lines[len(lines)-1] + "\\n" // add the "\n" back to the last line
```
VladislavsPerkanuks, line 173:
```suggestion
		lines[len(lines)-1] += "\\n" // add the "\n" back to the last line
```

pkg/convert/pot.go
```diff
@@ -110,47 +74,209 @@ func FromPot(b []byte) (model.Messages, error) {
...
+
+	return nil
+}
+
+func writeTags(b *bytes.Buffer, message model.Message) error {
+	if err := writeToPoTag(b, MsgId, message.ID); err != nil {
+		return fmt.Errorf("format msgid: %w", err)
+	}
+
+	if message.PluralID != "" {
```
VladislavsPerkanuks, line 267: `switch message.PluralID != ""` should be more readable than so many if's

pkg/pot/ast.go
```diff
@@ -46,6 +46,9 @@ func TokensToPo(tokens []Token) (Po, error) {
 	header := headerNode{}
 
 	for _, token := range tokens {
+		token.Value = strings.ReplaceAll(token.Value, "\\n ", "\n")
+		token.Value = strings.ReplaceAll(token.Value, "\\n", "\n")
```
VladislavsPerkanuks, line 50:
```suggestion
replacer := strings.NewReplacer("\\n ", "\n", "\\n", "\n")
token.Value = replacer.Replace(token.Value)
```
Same thing, but intend should be clearer

pkg/convert/pot.go
janishorsts: Inside `for` loop I recommend `continue`, it keeps the flow more linear.

```go
if len(pluralLines) == 1 {
  // ...
  continue
}
```

pkg/convert/pot.go
janishorsts: always

janishorsts (changes requested):
We need to take into account
https://docs.google.com/presentation/d/1ZyN8-0VXmod5hbHveq-M1AeQ61Ga3BmVuahZjbmbBxo/pub?start=false&loop=false&delayms=3000&slide=id.g1bc43a82_2_41

pkg/convert/pot.go
DarjaLeonova: Hey. I think in this function using a switch-case statement is not necessary, as there's only a single condition to check.
Also in this function, the if-else statement is more straightforward and easier to read.

pkg/convert/pot.go
```diff
@@ -110,47 +74,209 @@ func FromPot(b []byte) (model.Messages, error) {
...
+
+		switch {
+		case i == 0:
+			sb.WriteString(fmt.Sprintf("when 1 {%s}\n", line))
+		case i == len(plurals)-1:
+			sb.WriteString(fmt.Sprintf("when * {%s}\n", line))
+		default:
+			sb.WriteString(fmt.Sprintf("when %d {%s}\n", i+1, line))
+		}
+	}
```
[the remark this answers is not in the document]
DarjaLeonova: I think the code with 2 cases and 1 default is more readable because it presents conditions more explicitly.

pkg/convert/pot.go
DarjaLeonova: In this case, the if-else statement is more readable since there are only two branches.

VladislavsPerkanuks: I just dont like these nested if
```go
if ... {
  if ...{
  }
}
```

This adds variety to it ?
```go
switch ...{
  case ...:
    if...{
  }
}
```
e.g.:
```go
    switch len(lines) {
    case 1:
        if _, err = fmt.Fprintf(b, "%s \"%s\"\n", tag, lines[0]); err != nil {
            return fmt.Errorf("write %s: %w", tag, err)
        }
    default:
        if _, err = fmt.Fprintf(b, "%s \"\"\n", tag); err != nil {
            return fmt.Errorf("write %s: %w", tag, err)
        }
        if err = writeMultiline(b, tag, lines); err != nil {
            return fmt.Errorf("write multiline: %w", err)
        }
    }
```

But that is not so important :)

pkg/convert/pot.go
VladislavsPerkanuks: Nevertheless, DRY
```go
func writeMultiline(b *bytes.Buffer, tag poTag, lines []string) error {
    for _, line := range lines {
        if !strings.HasSuffix(line, "\\n") {
            line += "\\n"
        }
        if _, err := fmt.Fprint(b, "\""+line+"\""+"\n"); err != nil {
            return fmt.Errorf("write %s: %w", tag, err)
        }
    }
    return nil
}
```

pkg/convert/pot.go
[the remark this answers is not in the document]
VladislavsPerkanuks: Maybe you are right, but I just don't like repetition because here, `case i==0` and `default` does the same thing

Nevertheless `switch i {}` needs to be used here

pkg/convert/pot.go
```diff
@@ -73,46 +51,124 @@ func FromPot(b []byte) (model.Messages, error) {
...
+		message := model.Message{ID: node.MsgId}
+
+		switch {
+		case po.Header.PluralForms.NPlurals > pluralCountLimit:
+			return model.Messages{}, fmt.Errorf("plural forms with more than 2 forms are not implemented yet")
+		case po.Header.PluralForms.NPlurals > 1:
+			message.PluralID = node.MsgIdPlural
+			message.Message = convertToFormatMessage(node.MsgStr)
+		default:
+			message.Message = node.MsgStr[0]
```
VladislavsPerkanuks, line 65:
```suggestion
	for _, node := range po.Messages {
		message := model.Message{ID: node.MsgId, PluralID: node.MsgIdPlural}

		switch {
		case po.Header.PluralForms.NPlurals > pluralCountLimit:
			return model.Messages{}, fmt.Errorf("plural forms with more than 2 forms are not implemented yet")
		case po.Header.PluralForms.NPlurals > 1:
			message.Message = convertToFormatMessage(node.MsgStr)
		default:
			message.Message = node.MsgStr[0]
		}
```

You could define message with on 56 line with `MsgIDPLural`, if I am not mistaken when `po.Header.PluralForms.NPlurals` is 1, then `MsgIDPLural` is empty string.
If not nevermind

pkg/convert/pot.go
```diff
@@ -121,36 +177,116 @@ func writeToPoTag(b *bytes.Buffer, tag poTag, str string) error {
...
+		if _, err := fmt.Fprint(b, "#, fuzzy\n"); err != nil {
+			return fmt.Errorf("write fuzzy: %w", err)
+		}
+	}
+
+	if err := writeTags(b, message); err != nil {
+		return err
+	}
+
+	return nil
```
VladislavsPerkanuks, line 268:
```suggestion
	return writeTags(b, message)
```

VladislavsPerkanuks: try to be consistent with error messages, use Present Simple everywhere in errors
e.g
```go
fmt.Errorf("divide ...")
fmt.Errorf("convert ...")
```

pkg/pot/ast.go
```diff
@@ -41,20 +41,20 @@ type Po struct {
...
 		switch token.Type {
 		case HeaderLanguage:
-			parts := strings.Split(token.Value, ":")
-			if len(parts) < partsN {
-				return Po{}, fmt.Errorf("invalid language header format")
-			}
-
-			languageCode := strings.TrimSpace(parts[1])
-			header.Language = language.Make(languageCode)
+			header.Language = language.Make(token.Value)
```
VladislavsPerkanuks, line 57: `language.Parse` is superior for data integrity

pkg/convert/pot.go
```diff
@@ -4,64 +4,42 @@ import (
...
 
 	return b.Bytes(), nil
 }
 
 func FromPot(b []byte) (model.Messages, error) {
+	pluralCountLimit := 2
+
 	tokens, err := pot.Lex(bytes.NewReader(b))
 	if err != nil {
 		return model.Messages{}, fmt.Errorf("dividing po file to tokens: %w", err)
```
VladislavsPerkanuks, line 45: try to be consistent with error messages, use Present Simple everywhere in errors
e.g

fmt.Errorf("divide ...")
fmt.Errorf("convert ...")

VladislavsPerkanuks:
Could you explain, how the PluralID works?
In our current implementation, in messages we have
ID: Untranslated message
Message: Translated message

But you added another field of PluralID, and I kinda don't understand the reason behind that.

I was thinking about plurals and came up with the following:
```go
type Message struct {
	Original    string
	Translated  text
	Description string
	Fuzzy       bool
}

type text struct {
	Plural struct {
		Zero, One, Two, Few, Many, Other string
	}
	Singular string
}
```

But it seems complex to implement.

pkg/convert/pot.go
```diff
@@ -73,46 +53,175 @@ func FromPot(b []byte) (model.Messages, error) {
...
+		case po.Header.PluralForms.NPlurals > 1:
+			message.Message = convertToFormatMessage(node.MsgStr)
+		default:
+			message.Message = node.MsgStr[0]
 		}
 
-		if node.MsgIdPlural != "" {
-			continue
+		message.Description = strings.Join(node.ExtractedComment, "\n ")
+		message.Fuzzy = strings.Contains(node.Flag, "fuzzy")
```
VladislavsPerkanuks, line 70: could be combined with declaration on line 58

pkg/convert/pot.go
```diff
@@ -73,46 +53,175 @@ func FromPot(b []byte) (model.Messages, error) {
...
+			return fmt.Errorf("write %s: %w", tag, err)
+		}
+	}
+
+	return nil
+}
+
+func getPoTagLines(str string, tag poTag) ([]string, error) {
 	encodedStr, err := json.Marshal(str) // use JSON encoding to escape special characters
 	if err != nil {
```
VladislavsPerkanuks, line 223: how about [strconv.Qoute](https://pkg.go.dev/strconv#Quote) to remove special characters and minimize overhead
```suggestion
func getPoTagLines(str string, tag poTag) ([]string, error) {
    encodedStr := strconv.Quote(str)
```

pkg/convert/pot_test.go
```diff
@@ -49,7 +51,7 @@ msgstr "Au revoir!"
 `),
 		},
 		{
-			name: "When description is multiline",
+			name: "description is multiline",
```
VladislavsPerkanuks, line 54: I think it could be shortened even more:
```suggestion
			name: "multiline description",
```

and for all others test names too

pkg/convert/pot_test.go
```diff
@@ -114,7 +120,7 @@ msgstr "Au revoir!"
 `),
 		},
 		{
-			name: "When msgid is multiline but with single line",
+			name: "msgid is multiline but with single line",
```
VladislavsPerkanuks, line 123:
```suggestion
			name: "single msgid with newline",
```
more concise == better

janishorsts (changes requested):

pkg/convert/pot.go
```diff
@@ -73,46 +53,175 @@ func FromPot(b []byte) (model.Messages, error) {
...
+
 	for _, node := range po.Messages {
-		if strings.HasSuffix(node.MsgId, "\\n") {
-			node.MsgId = strings.ReplaceAll(node.MsgId, "\\n", "\n")
+		message := model.Message{ID: node.MsgId, PluralID: node.MsgIdPlural}
+
+		switch {
+		case po.Header.PluralForms.NPlurals > pluralCountLimit:
+			return model.Messages{}, errors.New("plural forms with more than 2 forms are not implemented yet")
+		case po.Header.PluralForms.NPlurals > 1:
```
janishorsts, line 63:
```suggestion
		case po.Header.PluralForms.NPlurals == pluralCountLimit:
```

pkg/convert/pot.go
```diff
@@ -73,46 +53,175 @@ func FromPot(b []byte) (model.Messages, error) {
 	}
 
 	messages := make([]model.Message, 0, len(po.Messages))
-	// model.Messages does not support plurals atm. But we plan to impl it under:
-	// https://github.com/orgs/translate-agent/projects/1?pane=issue&itemId=21251425
+
 	for _, node := range po.Messages {
-		if strings.HasSuffix(node.MsgId, "\\n") {
-			node.MsgId = strings.ReplaceAll(node.MsgId, "\\n", "\n")
+		message := model.Message{ID: node.MsgId, PluralID: node.MsgIdPlural}
```
janishorsts, line 58: Declare and initialize all values at the same time. As @VladislavsPerkanuks already mentioned.

```suggestion
		message := model.Message{
			ID: node.MsgId, 
			PluralID: node.MsgIdPlural
			Description: strings.Join(node.ExtractedComment, "\n "),
			Fuzzy: strings.Contains(node.Flag, "fuzzy"),
		}
```

janishorsts, line 58: Declare and initialize all values at the same time. As @VladislavsPerkanuks already mentioned.

```suggestion
		message := model.Message{
			ID: node.MsgId, 
			PluralID: node.MsgIdPlural
			Description: strings.Join(node.ExtractedComment, "\n "),
			Fuzzy: strings.Contains(node.Flag, "fuzzy"),
		}
```

pkg/convert/pot.go
```diff
@@ -73,46 +53,175 @@ func FromPot(b []byte) (model.Messages, error) {
...
+
+func writeToPoTag(b *bytes.Buffer, tag poTag, str string) error {
+	var write func(*bytes.Buffer, poTag, string) error
+
+	switch tag { //nolint:exhaustive
+	case MsgStrPlural:
+		write = writePluralForms
+	default:
+		write = writeDefault
+	}
```
janishorsts, line 89: We can rename `writePluralForms` to `writePlural` and `writeDefault` to `writeSingular` to use the same language and context.

Also, this makes it short and succinct.

```suggestion
	write := writeSingular

	if tag == MsgStrPlural {
		write = writePlural
	}
```

pkg/convert/pot.go
```diff
@@ -2,69 +2,47 @@ package convert
...
+		if err := writeMessage(&b, i, message); err != nil {
+			return nil, fmt.Errorf("write message: %w", err)
 		}
 	}
 
 	return b.Bytes(), nil
 }
 
 func FromPot(b []byte) (model.Messages, error) {
+	pluralCountLimit := 2
```
janishorsts, line 41:
```suggestion
	const pluralCountLimit = 2
```

PR #47, opened 2023-03-28
40 Repo implementation in service handlers

VladislavsPerkanuks:
Changes:
- Implement CRUD operations for the services in the server handlers
- Added integration tests for `Services` endpoints 
- Added `CreateService` in proto
- Updated proto definition with REST methods
- Changed from `cmux.multiplexer` to `grpc gateway` multiplexer, to fix PATCH request

### DIFF
--- a/pkg/translate/service.go
+++ b/pkg/translate/service.go
@@ -143,6 +143,10 @@ func (u *updateServiceParams) validate() error {
 		return errors.New("'service' is required")
 	}
 
+	if u.service.ID == uuid.Nil {
+		return errors.New("'service.id' is required")
+	}
+
 	if u.mask != nil {
 		for _, path := range u.mask.Paths {
 			if !slices.Contains(updateMaskAcceptablePaths, path) {
--- a/pkg/translate/service_test.go
+++ b/pkg/translate/service_test.go
@@ -236,14 +236,15 @@ func Test_ValidateUpdateServiceParams(t *testing.T) {
 
 	happyParams := randParams()
 
-	happyParamsMissingServiceID := randParams()
-	happyParamsMissingServiceID.service.ID = uuid.Nil
-
 	happyParamsMissingUpdateMask := randParams()
 	happyParamsMissingUpdateMask.mask = nil
 
 	missingServiceParams := randParams()
 	missingServiceParams.service = nil
+
+	// when updating a service, the service.ID is required and validation fails without it.
+	missingServiceIDParams := randParams()
+	missingServiceIDParams.service.ID = uuid.Nil
 
 	invalidUpdateMaskPathParams := randParams()
 	invalidUpdateMaskPathParams.mask.Paths = gofakeit.NiceColors()
@@ -259,11 +260,6 @@ func Test_ValidateUpdateServiceParams(t *testing.T) {
 			expectedErr: nil,
 		},
 		{
-			name:        "Happy Path Missing Service ID",
-			params:      happyParamsMissingServiceID,
-			expectedErr: nil,
-		},
-		{
 			name:        "Happy Path Missing Update Mask",
 			params:      happyParamsMissingUpdateMask,
 			expectedErr: nil,
@@ -272,6 +268,11 @@ func Test_ValidateUpdateServiceParams(t *testing.T) {
 			name:        "Missing Service",
 			params:      missingServiceParams,
 			expectedErr: errors.New("'service' is required"),
+		},
+		{
+			name:        "Missing Service ID",
+			params:      missingServiceIDParams,
+			expectedErr: errors.New("'service.id' is required"),
 		},
 		{
 			name:        "Invalid Update Mask Path",
@@ -560,6 +561,10 @@ func Test_ValidateCreateServiceParams(t *testing.T) {
 
 	happyParams := randParams()
 
+	// when creating a service, the service.ID is optional and validation passes.
+	happyParamsEmptyServiceId := randParams()
+	happyParamsEmptyServiceId.service.ID = uuid.Nil
+
 	emptyServiceParams := randParams()
 	emptyServiceParams.service = nil
 
@@ -571,6 +576,11 @@ func Test_ValidateCreateServiceParams(t *testing.T) {
 		{
 			name:        "Happy Path",
 			params:      happyParams,
+			expectedErr: nil,
+		},
+		{
+			name:        "Happy Path Empty Service ID",
+			params:      happyParamsEmptyServiceId,
 			expectedErr: nil,
 		},
 		{


cmd/translate/main_grpc_integration_test.go
```diff
@@ -201,3 +204,175 @@ func Test_DownloadTranslationFile_gRPC(t *testing.T) {
...
+	_, err := client.CreateService(ctx, &translatev1.CreateServiceRequest{Service: service})
+	if !assert.NoError(t, err, "Prepare test data") {
+		return
+	}
+
+	service.Name = gofakeit.FirstName()
+
+	_, err = client.UpdateService(ctx, &translatev1.UpdateServiceRequest{
+		Service:    service,
+		UpdateMask: &fieldmaskpb.FieldMask{Paths: []string{"name"}},
```
VladislavsPerkanuks, line 277: When adding FieldMask, should it be `service.name`, or just `service` ?

klavinsjanis: for primary resource "name", for nested subresource.name

pkg/translate/service.go
```diff
@@ -9,16 +9,25 @@ import (
...
 )
 
 type (
 	getServiceRequest    translatev1.GetServiceRequest
+	createServiceRequest translatev1.CreateServiceRequest
 	updateServiceRequest translatev1.UpdateServiceRequest
 	deleteServiceRequest translatev1.DeleteServiceRequest
 )
 
+// protoServiceFromService converts model.Service -> translatev1.Service.
```
klavinsjanis, line 26: maybe better to use to instead of ->

cmd/translate/main.go
```diff
@@ -125,11 +128,13 @@ func init() {
 	rootCmd.PersistentFlags().StringVar(&cfgFile, "config", "translate.yaml", "config file")
 	rootCmd.PersistentFlags().Uint("port", 8080, "port to run service on") //nolint:gomnd
 	rootCmd.PersistentFlags().String("host", "localhost", "host to run service on")
+	rootCmd.PersistentFlags().String("db", "mysql", "db to use with service")
```
klavinsjanis, line 131: "database to use ...

.vscode/launch.json
```diff
@@ -0,0 +1,20 @@
...
+    {
+      "name": "Launch Package",
+      "type": "go",
+      "request": "launch",
+      "mode": "auto",
+      "program": "${workspaceFolder}/cmd/translate/main.go",
+      "cwd": "${workspaceFolder}",
+      "envFile": "${workspaceFolder}/.env.test",
+      "env": {
+        "OTEL_RESOURCE_ATTRIBUTES": "deployment.environment=debug"
```
janishorsts, line 16: Do not override existing `deployment.environment`!

.vscode/launch.json
```diff
@@ -0,0 +1,20 @@
+{
+  // Use IntelliSense to learn about possible attributes.
+  // Hover to view descriptions of existing attributes.
+  // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+  "version": "0.2.0",
+  "configurations": [
+    {
+      "name": "Launch Package",
```
janishorsts, line 8: ?

```suggestion
      "name": "Debug translate service",
```

VladislavsPerkanuks: Yeah, will, this was the default for generating launch.json

cmd/translate/main.go
```diff
@@ -57,10 +71,27 @@ var rootCmd = &cobra.Command{
 		defer grpcServer.GracefulStop()
 
 		mux := runtime.NewServeMux()
-		translatev1.RegisterTranslateServiceServer(grpcServer, &translate.TranslateServiceServer{})
+
+		var (
+			repository repo.Repo
+			errRepo    error
```
janishorsts, line 77: Advisory. In this case, you don't have to declare a new error instance. You can reuse `err`.

cmd/translate/main_grpc_integration_test.go
```diff
@@ -201,3 +204,179 @@ func Test_DownloadTranslationFile_gRPC(t *testing.T) {
...
+
+	_, err := client.CreateService(context.Background(), &translatev1.CreateServiceRequest{Service: service})
+	if !assert.NoError(t, err, "Prepare test data") {
+		return
+	}
+
+	actual := status.Code(err)
+	expected := codes.OK
+
+	assert.Equal(t, expected, actual, "want codes.%s got codes.%s", expected, actual)
```
janishorsts, line 230: Up to you. But to reduce code boilerplate code (tests by nature are very noisy). It makes it easier to read the tests.

```suggestion
	assert.Equal(t, codes.OK, status.Code(err))
```

VladislavsPerkanuks: I would agree with you, but when tests fail with 
```go
assert.Equal(t, codes.OK, status.Code(err)) 
```
the expected and actual status codes are not human readable, while 
```go
assert.Equal(t, expected, actual, "want codes.%s got codes.%s", expected, actual)
```
On fail prints human-readable expected and actual status code, so I will stick with latter

pkg/transform/service.go
```diff
@@ -0,0 +1,11 @@
+package transform
+
+import (
+	"go.expect.digital/translate/pkg/model"
+	translatev1 "go.expect.digital/translate/pkg/pb/translate/v1"
+)
+
+// ProtoServiceFromService converts model.Service to translatev1.Service.
+func ProtoServiceFromService(service *model.Service) *translatev1.Service {
```
janishorsts, line 9: Keep the naming simple. E.g.

```suggestion
func ServiceFromPb(service *translatev1.Service) *model.Service
func ServiceToPb(service *model.Service) *translatev1.Service

// or

func ServiceFromProto(service *translatev1.Service) *model.Service
func ServiceToProto(service *model.Service) *translatev1.Service
```

pkg/transform/service.go
```diff
@@ -0,0 +1,11 @@
...
+	return &translatev1.Service{Id: service.ID.String(), Name: service.Name}
```
janishorsts, line 10: if uuid.UUID is empty (uuid.Nil), `.String()` returns '00000000-0000-0000-0000-000000000000' but we expect empty string value in the proto message. Set only if it's not uuid.Nil.

pkg/transform/service.go
janishorsts, line 1: Use property testing for the model transformations

```go
quick.Check(func (id uuid.UUID, name string) {
  service := &model.Service{ID: id, Name: name}

  return assert.Equal(t, service, ServiceFromPb(ServiceToPb(service)))
}, nil)
```

pkg/translate/service.go
```diff
@@ -68,12 +76,78 @@ func (t *TranslateServiceServer) ListServices(
...
+	services, err := t.repo.LoadServices(ctx)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, err.Error())
+	}
+
+	resp := &translatev1.ListServicesResponse{
+		Services: make([]*translatev1.Service, 0, len(services)),
+	}
+
+	for _, v := range services {
```
janishorsts, line 88: We can use a generic function for the slice transformation. Please check how it is done in the `tw` project - `transform/transform.go`.

pkg/translate/transform.go
VladislavsPerkanuks, line 1: Moved from a separate package to translate pkg.
**Reason**: Only translate pkg needs access to transform (pb -> ours), so there is no need to create separate pkg and create exported functions

pkg/translate/service_test.go
```diff
@@ -4,38 +4,49 @@ import (
...
+	}
+}
+
 func Test_ParseGetServiceParams(t *testing.T) {
 	t.Parallel()
 
+	happyReq := randGetServiceReq()
+
+	malformedIDReq := randGetServiceReq()
+	malformedIDReq.Id += "_FAIL" //nolint:goconst
```
VladislavsPerkanuks, line 30: Switched to that kind of declaration for test inputs

- Table tests are shorter and more concise
- Random data is generated, which could potentially point to some uncovered edge case

pkg/translate/transform.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	return id, nil
+}
+
+// ----------------------Service----------------------
+
+// serviceToProto converts model.Service to translatev1.Service.
+func serviceToProto(s *model.Service) *translatev1.Service {
+	return &translatev1.Service{Id: uuidToProto(s.ID), Name: s.Name}
```
janishorsts, line 40: Cover the case when s is nil when transforming models.

```suggestion
	if s == nil {
		return nil
	}
	
	return &translatev1.Service{Id: uuidToProto(s.ID), Name: s.Name}
```

pkg/translate/transform.go
```diff
@@ -0,0 +1,91 @@
...
+	service.ID, err = uuidFromProto(s.Id)
+	if err != nil {
+		return nil, fmt.Errorf("transform id: %w", err)
+	}
+
+	return service, nil
+}
+
+// servicesToProto converts []model.Service to []*translatev1.Service.
+func servicesToProto(s []model.Service) []*translatev1.Service {
```
janishorsts, line 59: Use a generic function to transform a slice. E.g.

```suggestion
func servicesToProto(s []model.Service) []*translatev1.Service { return sliceToProto(s) }

func sliceToProto[T any, R any](slice []T, transform func(*T) *R) []R {
	if len(slice) == 0 {
		return nil
	}

	v := make([]R, 0, len(slice))

	for i := range slice {
		v = append(v, *transform(&slice[i]))
	}

	return v
}
```

pkg/translate/transform_test.go
```diff
@@ -0,0 +1,76 @@
...
+	t.Run("UUID to string to UUID", func(t *testing.T) {
+		t.Parallel()
+
+		f := func(expectedID uuid.UUID) bool {
+			stringID := uuidToProto(expectedID)
+
+			restoredID, err := uuidFromProto(stringID)
+			require.NoError(t, err)
+
+			return assert.Equal(t, expectedID, restoredID)
```
janishorsts, line 25:
```suggestion

			return assert.NoError(t, err) && assert.Equal(t, expectedID, restoredID)
```

pkg/translate/service.go
```diff
@@ -68,35 +70,96 @@ func (t *TranslateServiceServer) ListServices(
...
+	return &createServiceParams{service: service}, nil
+}
+
+func (t *TranslateServiceServer) CreateService(
+	ctx context.Context,
+	req *translatev1.CreateServiceRequest,
+) (*translatev1.Service, error) {
+	createReq := (*createServiceRequest)(req)
+
+	params, err := createReq.parseParams()
```
janishorsts, line 110: Let's simplify this abstraction - it does not add anything. Consider using functional programming paradigms first.

Please apply it to all request parsing.

```suggestion
	params, err := parseCreateServiceRequest(req)
```

janishorsts (changes requested):
Please be careful what error messages are exposed to the gRPC clients. We do not want to expose internal server errors etc.

PS. I added a ticket to instrument traces in integration tests https://github.com/orgs/translate-agent/projects/1/views/1?pane=issue&itemId=25756421

pkg/translate/service.go
```diff
@@ -152,9 +224,12 @@ func (t *TranslateServiceServer) DeleteService(
 		return nil, status.Errorf(codes.InvalidArgument, err.Error())
 	}
 
-	if err := params.validate(); err != nil {
-		return nil, status.Errorf(codes.InvalidArgument, err.Error())
+	switch err := t.repo.DeleteService(ctx, params.id); {
+	default:
+		return &emptypb.Empty{}, nil
```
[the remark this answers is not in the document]
VladislavsPerkanuks: Cannot do it. It breaks gRPC

image

pkg/translate/service.go
```diff
@@ -25,41 +23,44 @@ type getServiceParams struct {
...
+		return nil, fmt.Errorf("parse id: %w", err)
 	}
 
-	return getServiceParams{id: id}, nil
+	return &getServiceParams{id: id}, nil
 }
 
-func (g *getServiceParams) validate() error {
-	// validate if uuid is in DB
+func validateGetServiceRequestParams(params *getServiceParams) error {
```
janishorsts, line 35: We can leave it as it was previously - `func (g *getServiceParams) validate() error`.

The benefit of `parseGetServiceRequestParams` is it removes the need for the declaration of `params` in a consumer function, and keeps the consumer function tidier.

janishorsts (approved):
Created a new ticket to validate field mask https://github.com/orgs/translate-agent/projects/1/views/1?pane=issue&itemId=26154387

PR #52, opened 2023-04-03
Translation File Upload/Download Implementation

VladislavsPerkanuks:
This PR implements the Save and Download Translation File feature.

## Changes

- ***Repo***:
  - A new tables `message` and `message_message` has been added to the migrate.
  - The Save and Load Messages feature has been implemented.
  - Integration tests for the repo Save/Load Messages has been added.
- ***Server***:
  - Server handlers for Save/Download Translation file have been implemented.
  - Integration tests for these handlers have been added.
  - The REST path for download/upload has been updated to `/v1/services/{service_id}/files/{language}`, GET and PUT respectively.
  - The params parser in download/upload handlers has been updated.

~~***Note***: `CI/CD Test` will fail due to POT conversion panicking (model.Messages -> POT serialized data -> model.Messages). Waiting for the fix, to pass checks.~~
**Update** `POT` is omitted when generating rand test data, to avoid panic

PR #54, opened 2023-04-11
51-upload-file-using-translate-client

klavinsjanis:


cmd/translate/main_client_integration_test.go
```diff
@@ -34,3 +36,117 @@ func Test_ServiceLs(t *testing.T) {
 		assert.Nil(t, res)
 	})
 }
+
+func Test_ServiceUploadCmd(t *testing.T) {
+	t.Run("OK", func(t *testing.T) {
+		file, err := os.CreateTemp(t.TempDir(), "test")
```
VladislavsPerkanuks, line 42: Please close and delete the file, to avoid a mess
```go
defer file.Close()
defer os.Remove(file.Name())
```

VladislavsPerkanuks (commented):
Before merging into main, you should wait for #52 to be merged, as there are some changes

cmd/translate/main_client_integration_test.go
VladislavsPerkanuks: > It is the caller's responsibility to remove the file when it is no longer needed.

https://pkg.go.dev/os#CreateTemp

VladislavsPerkanuks: I guess you are right, `t.TempDir` gets removed, after tests :)

cmd/client/cmd/service.go
```diff
@@ -2,9 +2,11 @@ package cmd
...
 import (
 	"context"
+	"errors"
 	"fmt"
 	"log"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/jedib0t/go-pretty/v6/table"
```
janishorsts, line 12: Let's use this one instead https://github.com/rodaine/table . It's simpler and more compact output.

cmd/client/cmd/service.go
```diff
@@ -90,6 +92,84 @@ func newLsCmd() *cobra.Command {
...
+func newUploadCmd() *cobra.Command {
+	var schemaFlag schema
+
+	uploadCmd := &cobra.Command{
+		Use:   "upload",
+		Short: "Upload file to translate agent service",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			timeout, err := cmd.InheritedFlags().GetDuration("timeout")
+			if err != nil {
+				return fmt.Errorf("upload file: retrieve cli parameter 'timeout': %w", err)
```
janishorsts, line 104:
```suggestion
				return fmt.Errorf("upload file: get cli parameter 'timeout': %w", err)
```

cmd/client/cmd/service.go
```diff
@@ -90,6 +92,84 @@ func newLsCmd() *cobra.Command {
...
+			defer cancelFunc()
+
+			client, err := newClientConn(ctx, cmd)
+			if err != nil {
+				return fmt.Errorf("upload file: new GRPC client connection: %w", err)
+			}
+
+			language, err := cmd.Flags().GetString("language")
+			if err != nil {
+				return fmt.Errorf("upload file: retrieve cli parameter 'language': %w", err)
```
janishorsts, line 117:
```suggestion
				return fmt.Errorf("upload file: get cli parameter 'language': %w", err)
```

cmd/client/cmd/service.go
```diff
@@ -90,6 +92,84 @@ func newLsCmd() *cobra.Command {
...
+			if _, err = fmt.Fprintln(cmd.OutOrStdout(), "File uploaded successfully."); err != nil {
+				return fmt.Errorf("upload file: output response to stdout: %w", err)
+			}
+
+			return nil
+		},
+	}
+
+	uploadFlags := uploadCmd.Flags()
+	uploadFlags.StringP("path", "p", "", "file path")
```
janishorsts, line 150:
```suggestion
	uploadFlags.StringP("file", "f", "", "file path")
```

janishorsts (approved):
Replace the table package. Everything else is a minor issue. And has a very high probability of changing in the future.

PR #56, opened 2023-04-13
46# parse message format v2

DarjaLeonova:
Parser for message format v2 currently supports only plurals.


pkg/messageFormat/ast.go
```diff
@@ -0,0 +1,101 @@
+package messageFormat
```
klavinsjanis, line 1: Good package names are short and clear. They are lower case, with no under_scores or mixedCaps.

pkg/messageFormat/ast.go
```diff
@@ -0,0 +1,101 @@
...
+	)
+
+	// If message format starts with placeholder {
+	if tokens[0].Type == PlaceholderOpen {
+		tokens[0] = emptyToken
+		tokens[len(tokens)-1] = emptyToken
+	}
+
+	for i, token := range tokens {
+		switch token.Type { //nolint:exhaustive
```
klavinsjanis, line 56: add a case with all the missing types to avoid the use of nolint:exhaustive
case x, y, z:
// noop

pkg/messageFormat/ast.go
```diff
@@ -0,0 +1,101 @@
...
+	Keys    []string
+	Message []interface{}
+}
+
+type NodeFunction struct {
+	Name string
+}
+
+func TokensToMessageFormat(tokens []Token) (NodeMatch, error) {
+	var (
```
klavinsjanis, line 40: what happens if tokens slice is empty? Maybe return an error?

pkg/messageFormat/ast_test.go
```diff
@@ -0,0 +1,299 @@
...
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			result, err := TokensToMessageFormat(tt.input)
+
+			if tt.expectedErr != nil {
+				assert.Errorf(t, err, tt.expectedErr.Error())
+				return
+			}
+
+			if !assert.NoError(t, err) {
```
klavinsjanis, line 292: shorter version:
require.NoError(t, err)

pkg/messageFormat/lexer.go
```diff
@@ -0,0 +1,129 @@
...
+
+func Lex(str string) ([]Token, error) {
+	var (
+		tokens           []Token
+		parsedTokens     []Token
+		buf              []rune
+		placeholderLevel int
+	)
+
+	for i, r := range str {
```
klavinsjanis, line 34: Note that index won't always represent the position of the character in string, this might only work for ASCII characters. In Go Strings are UTF-8 encoded, this means each charcter called rune can be of 1 to 4 bytes long.

for pos, char := range "日本語" {
    fmt.Printf("character %c starts at byte position %d\n", char, pos)
}
This prints :
character 日 starts at byte position 0
character 本 starts at byte position 3
character 語 starts at byte position 6

pkg/messageFormat/lexer.go
```diff
@@ -0,0 +1,129 @@
...
+	if len(buf) > 0 {
+		tokenizeBuffer(buf, &tokens, &placeholderLevel)
+	}
+
+	parsedTokens = combineTextTokens(tokens, parsedTokens)
+
+	return parsedTokens, nil
+}
+
+func combineTextTokens(tokens []Token, parsedTokens []Token) []Token {
```
klavinsjanis, line 86: combineTextTokens(token, parsedTokens []Token)

pkg/messageFormat/lexer.go
```diff
@@ -0,0 +1,129 @@
...
+	txt := ""
+
+	for y := range tokens {
```
klavinsjanis, line 89: index variable is usually represented with letter i

pkg/messageFormat/ast.go
```diff
@@ -0,0 +1,101 @@
+package messageFormat
+
+import (
+	"fmt"
+	"strings"
+)
+
+type NodeMatch struct {
```
klavinsjanis, line 8: Note: that types that are not used or planned to be used outside the package shouldn't be exposed.

pkg/message/lexer.go
```diff
@@ -0,0 +1,139 @@
...
+			}
+		} else {
+			parsedTokens = append(parsedTokens, tokens[i])
+		}
+	}
+
+	return parsedTokens
+}
+
+func tokenizeBuffer(buf []rune, tokens *[]Token, placeholderLevel *int) {
```
klavinsjanis, line 115: tokenizeBuffer? Please add function descriptions.

pkg/message/lexer_test.go
VladislavsPerkanuks, line 1: I would suggest using random data, with `gofakeit`, that would help find some edge cases, if there is any

pkg/message/ast.go
```diff
@@ -0,0 +1,101 @@
...
+	var (
+		match               nodeMatch
+		currentVariant      nodeVariant
+		currentExpr         nodeExpr
+		isPlaceholderClosed bool
+		currentLevel        int
+		token               Token
+	)
+
+	if len(tokens) == 0 {
```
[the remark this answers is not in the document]
VladislavsPerkanuks: agree

pkg/message/ast_test.go
```diff
@@ -0,0 +1,304 @@
...
+							nodeText{
+								Text: " notifications.",
+							},
+						},
+					},
+				},
+			},
+		},
+		{
+			name: "message with multiple plurals and selctors",
```
VladislavsPerkanuks, line 185:
```suggestion
			name: "message with multiple plurals and selectors",
```

pkg/message/ast.go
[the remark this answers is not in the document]
VladislavsPerkanuks: agree

pkg/message/lexer_test.go
VladislavsPerkanuks: Can't help with that, I just suggested this as an option. If it is possible it would be nice to have otherwise nothing can be done :)

pkg/message/lexer.go
```diff
@@ -0,0 +1,161 @@
...
+	parsedTokens, err := combineTextTokens(tokens, parsedTokens)
+	if err != nil {
+		return []Token{}, errors.New("combine Text tokens")
+	}
+
+	return parsedTokens, nil
+}
+
+// Combining Text tokens into one sentence.
+func combineTextTokens(tokens, parsedTokens []Token) ([]Token, error) {
```
klavinsjanis, line 103: Function descriptions should start with the function name followed by the description.
example:
// combineTextTokens combines ...

pkg/message/ast.go
```diff
@@ -0,0 +1,101 @@
...
+type nodeVariant struct {
+	Keys    []string
+	Message []interface{}
+}
+
+type nodeFunction struct {
+	Name string
+}
+
+func TokensToMessageFormat(tokens []Token) (nodeMatch, error) {
```
janishorsts, line 35: Never return private types (nodeMatch) in exposed functions.

janishorsts, line 35:
```suggestion
func Parse(s string) (NodeMatch, error)

func tree(tokens []Token) (NodeMatch, error)
```

pkg/message/ast.go
```diff
@@ -0,0 +1,101 @@
+package message
+
+import (
+	"errors"
+	"strings"
+)
+
+type nodeMatch struct {
```
janishorsts, line 8: Expose all nodes.

pkg/message/ast.go
janishorsts, line 35: We'll do pair programming to build AST.

janishorsts (changes requested):
I guess you can delete the `message` package.

janishorsts (approved):

PR #59, opened 2023-04-17
55-download-file-using-translate-client

klavinsjanis:


cmd/client/cmd/root.go
```diff
@@ -0,0 +1,116 @@
...
+
+	if connIsInsecure {
+		opts = append(opts, grpc.WithTransportCredentials(insecure.NewCredentials()))
+	}
+
+	// Wait for the server to start and establish a connection.
+	return grpc.DialContext(ctx, address, opts...) //nolint:wrapcheck
+}
+
+type schema string
```
VladislavsPerkanuks, line 86: Couldn't you reuse in `pkg/pb/translate/v1/translate.pb.go` defined schema?

go.mod
```diff
@@ -26,15 +26,20 @@ require (
 require (
 	github.com/cenkalti/backoff/v4 v4.2.0 // indirect
 	github.com/davecgh/go-spew v1.1.1 // indirect
+	github.com/fatih/color v1.15.0
```
VladislavsPerkanuks, line 29: We are having direct imports in the first `require`, and indirect in second `require`

cmd/translate/main_client_integration_test.go
```diff
@@ -0,0 +1,316 @@
...
+			"service", "ls",
+			"-a", fmt.Sprintf("%s:%s", host, port),
+		})
+
+		assert.ErrorContains(t, err, "no transport security set")
+		assert.Nil(t, res)
+	})
+}
+
+func Test_ServiceUploadCmd(t *testing.T) {
```
VladislavsPerkanuks, line 43: MIssleading name.
TranslationFile Upload, not service upload

cmd/translate/main_client_integration_test.go
```diff
@@ -0,0 +1,316 @@
...
+			"-l", "xyz-ZY-Latn",
+			"-s", "json_ng_localize",
+		})
+
+		assert.ErrorContains(t, err, "required flag(s) \"serviceUUID\" not set")
+		assert.Nil(t, res)
+	})
+}
+
+func Test_ServiceDownloadCmd(t *testing.T) {
```
VladislavsPerkanuks, line 195: also misleading name

cmd/client/cmd/upload.go
```diff
@@ -0,0 +1,108 @@
...
+				return fmt.Errorf("upload file: output response to stdout: %w", err)
+			}
+
+			return nil
+		},
+	}
+
+	uploadFlags := uploadCmd.Flags()
+	uploadFlags.StringP("serviceUUID", "u", "", "service UUID")
+	uploadFlags.StringP("fileUUID", "p", "", "translation file UUID")
```
VladislavsPerkanuks, line 81: I think serviceID and fileID is more appropriate here

go.mod
klavinsjanis: wonder why they were placed there

cmd/client/cmd/root.go
klavinsjanis: I could use it if it had the proper methods.

VladislavsPerkanuks: I meant, something like this
```go
type schema translatev1.Schema

func (s schema) String() string{
  return translatev1.Schema_name[int32(s)]
}

...

```

I am just brainstorming, to avoid the repetition of code.

PR #62, opened 2023-04-25
Add Initial Instrumentation for Tracing in Integration Tests

VladislavsPerkanuks:
#61 
This is the first part of tracing implementation in integration tests. It covers only creating traces for tests.

## Changes
- Span creation has been added to integration tests
- Interceptors have been added for the grpc-gateway internal client
- The mux has been wrapped in otelhttp to enable tracing on REST methods
- Bumped OTEL to 1.15.1, and updated other go dependencies

| gRPC                                                                                                                                       | REST                                                                                                                                       |
|--------------------------------------------------------------------------------------------------------------------------------------------|--------------------------------------------------------------------------------------------------------------------------------------------|
| image | image |


cmd/translate/main_grpc_integration_test.go
```diff
@@ -35,6 +36,8 @@ var (
 	client translatev1.TranslateServiceClient
 )
 
+const name = "translate.server.integration_test"
```
janishorsts, line 39: I suggest declaring package level tracer and using it everywhere, reducing code boilerplate code.

The `name` in `otel.Tracer(name)` is stored in the attribute [otel.library.name](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/mapping-to-non-otlp.md#instrumentationscope) - deprecated and alias to `otel.scope.name`. [More about Instrumentation Scope ](https://opentelemetry.io/docs/specs/otel/glossary/#instrumentation-scope)

Based on the Instrumentation Scope description, I suggest the name `go.expect.digital/translate/cmd/translate`.

VladislavsPerkanuks: two things:
1. both must be defined otel.library.name + otel.scope.name
 > The following deprecated aliases MUST also be reported with exact same values for backward compatibility reasons:
2. Should we wait for OpenTelemetry to change this behavior in the Go package when creating a new tracer? While we can manually add a new attribute 'otel.scope.name' to traces that we create, all other instrumentation (such as gRPC, HTTP, and SQL) will still only have otel.library.name. That means inconsitency

cmd/translate/main_grpc_integration_test.go
```diff
@@ -157,7 +160,8 @@ func createService(ctx context.Context, t *testing.T) *translatev1.Service {
 func Test_UploadTranslationFile_gRPC(t *testing.T) {
 	t.Parallel()
 
-	ctx := context.Background()
+	ctx, span := otel.Tracer(name).Start(context.Background(), t.Name())
+	defer span.End()
```
janishorsts, line 164: It would be nice to mark the span as failed when the test fails to improve visibility.

```suggestion
func trace(ctx, t) (ctx, func()) {
  ctx, span := tracer.Start(ctx, t.name)
  
  return func() {
    if t.Failed() {
      span.SetStatus...
      span.RecordError...
    }
    
    span.End()
  }
}
	ctx, spanEnd := trace(ctx, t)
	defer spanEnd()
```

cmd/translate/main_grpc_integration_test.go
janishorsts: We do nothing about it - We don't add manually `otel.scope.name`, it is the responsibility of the OTEL go package. They are gonna do it when the time comes.

janishorsts: My understanding is, both are required to not break any behavior by collectors and GUIs. Otherwise, they are only recommended attributes.

cmd/translate/main_grpc_integration_test.go
```diff
@@ -157,102 +180,106 @@ func createService(ctx context.Context, t *testing.T) *translatev1.Service {
...
+		defer spanEnd()
 
-			actualCode := status.Code(err)
-			assert.Equal(t, tt.expectedCode, actualCode, "want codes.%s got codes.%s\nerr: %s", tt.expectedCode, actualCode, err)
-		})
-	}
-}
+		service := createService(prepareCtx, t)
 
-func Test_UploadTranslationFileDifferentLanguages_gRPC(t *testing.T) {
```
VladislavsPerkanuks, line 208: pointless test.
Already tested in repo integration tests

cmd/translate/main_grpc_integration_test.go
```diff
@@ -157,102 +180,106 @@ func createService(ctx context.Context, t *testing.T) *translatev1.Service {
...
 
-	for _, tt := range tests {
-		tt := tt
-		t.Run(tt.name, func(t *testing.T) {
-			t.Parallel()
+	var tests []test
 
-			_, err := client.UploadTranslationFile(ctx, tt.request)
+	// Prepare
+	t.Run("Prepare tests", func(t *testing.T) {
```
VladislavsPerkanuks, line 196: wrapping preparations for tests in subtest .e.g in this case inserting service for translation file, this way cleaner in Jaeger
image

cmd/translate/main_grpc_integration_test.go
```diff
@@ -400,52 +455,70 @@ func Test_CreateService_gRPC(t *testing.T) {
...
-	}{
-		{
-			name:         "Happy path all fields",
-			expectedCode: codes.OK,
-			request: &translatev1.UpdateServiceRequest{
-				Service:    randService(),
-				UpdateMask: nil,
+	type test struct {
+		request         *translatev1.UpdateServiceRequest
+		serviceToUpdate *translatev1.Service
```
VladislavsPerkanuks, line 463: Apparently, this test wasn't working as expected, fixed it

cmd/translate/main_grpc_integration_test.go
VladislavsPerkanuks: nevermind... it was bad idea :D

cmd/translate/main.go
```diff
@@ -15,6 +15,7 @@ import (
 	"github.com/spf13/cobra"
```
VladislavsPerkanuks, line 15: Use hide whitespace, to ease the review :)
image

klavinsjanis (approved):

janishorsts (approved):
One outstanding issue - integration subtests are not scoped, entities are created outside subtests resulting in the requests being outside traces.

PR #64, opened 2023-04-25
Detect Language from Uploaded File Body

VladislavsPerkanuks:
## Changes
- File language is now "optional" (see Possible Outcomes) when uploading translation file.
- Added an additional binding: `PUT v1/services/{service_id}/files`.
- Added an integration test for the new binding.
- Updated tests to pass with the new language parsing method.

### Possible Outcomes
1. If both the messages and request language are undefined, an error is returned.
2. If both are defined but not equal, an error is returned.
3. If the messages language is undefined and the request language is defined, the request language is returned.
4. If the messages language is defined and the request language is undefined, the messages language is returned.

### TODO
~~Waiting for #74 to be merged, into main so I could change, the CLI language parameter to optional~~


cmd/translate/main_grpc_integration_test.go
```diff
@@ -21,12 +21,19 @@ import (
 
 // -------------Translation File-------------.
 
-func randUploadData(t *testing.T, schema translatev1.Schema) ([]byte, language.Tag) {
+// TODO random model.Messages (and other model and proto structs/types) are used in multiple test files
+// Move rand data generation to separate pkg testutil/rand to avoid repetition ?
+// e.g. testutil/rand/proto.go, testutil/rand/model.go
+
+func randLang() language.Tag {
```
janishorsts, line 28: already defined - `rand.Language()`?

janishorsts (approved):

PR #66, opened 2023-04-26
Add Service Field Mask Validation

VladislavsPerkanuks:
## Changes:

- Added `model.Mask` with a type of `[]string`
- Added `parseFieldMask`, to parse and normalize paths and return `model.Mask`
- Updated tests


pkg/translate/field_mask_test.go
```diff
@@ -0,0 +1,92 @@
...
+	"testing"
+
+	"github.com/brianvoe/gofakeit/v6"
+	"github.com/stretchr/testify/assert"
+	translatev1 "go.expect.digital/translate/pkg/pb/translate/v1"
+	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/types/known/fieldmaskpb"
+)
+
+func Test_ValidateFieldMask(t *testing.T) {
```
VladislavsPerkanuks, line 14: Although these tests may seem excessive since a 3rd party package is being used and it likely has its own tests, they were originally implemented to loop through all fields of a proto message using reflection. Later on, it was discovered that the package has useful functions for this purpose. Despite this, the decision was made to keep the tests as they won’t cause any harm.

pkg/translate/service.go
```diff
@@ -122,20 +121,38 @@ func (t *TranslateServiceServer) CreateService(
 
 // ---------------------UpdateService-------------------------------
 
-var updateMaskAcceptablePaths = []string{"name"}
-
 type updateServiceParams struct {
 	mask    *fieldmaskpb.FieldMask
```
janishorsts, line 125: Please, add `model.Mask` []string. Otherwise, the pb struct leaks into the core system.

pkg/translate/field_mask.go
```diff
@@ -0,0 +1,24 @@
...
+import (
+	"fmt"
+
+	"go.expect.digital/translate/pkg/model"
+	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/types/known/fieldmaskpb"
+)
+
+// parseFieldMask parses and normalizes a field mask from a proto message and a list of paths.
+func parseFieldMask(message proto.Message, paths []string) (model.FieldMask, error) {
```
VladislavsPerkanuks, line 12: Future-proofing, when we will have multiple entities with field mask we won't be needing to call `fieldmaskpb.New` and `fieldmaskpb.Normalize` all over again

pkg/translate/service.go
```diff
@@ -6,11 +6,9 @@ import (
 	"fmt"
 
 	"github.com/google/uuid"
-	"golang.org/x/exp/slices"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/types/known/emptypb"
-	"google.golang.org/protobuf/types/known/fieldmaskpb"
 
 	"go.expect.digital/translate/pkg/model"
```
janishorsts, line 13:
```suggestion
	"go.expect.digital/translate/pkg/model"
```

janishorsts (approved):

PR #68, opened 2023-04-27
Update Service From Field Mask

VladislavsPerkanuks:
## Changes:
- Added a new function `updateFromMask` that updates a struct with values from another struct based on a field mask. The function uses custom `protoName` field tags to map field mask paths, to our Go model struct fields.
- Added a new function `updateServiceFromMask` that updates a Service with values from another Service based on a mask.
- Added tests for both
- Removed the old implementation and tests


pkg/translate/field_mask.go
```diff
@@ -22,3 +24,84 @@ func parseFieldMask(message proto.Message, paths []string) (model.Mask, error) {
...
+// updateModelFromFieldMask updates the dst with the values from src based on the fieldMask.
+//
+// Following scenarios are possible:
+//   - fieldMask is nil: All fields are updated.
+//   - fieldMask is not nil, but has no paths: No fields are updated.
+//   - fieldMask contains paths, that does not exist in the model: The paths are ignored.
+//   - fieldMask contains paths, that exist in model: Only those fields are updated.
+//
+// `protoName` tags are used to match fields from the fieldMask to fields in the model.
+func updateModelFromFieldMask[T any](mask model.Mask, dst, src *T) *T {
```
janishorsts, line 37: It's better if src is first followed by dst.

```suggestion
func copyMask[T any](src, dst *T, mask model.Mask) *T {
```

pkg/translate/field_mask_test.go
```diff
@@ -0,0 +1,282 @@
...
+	"testing"
+
+	"github.com/brianvoe/gofakeit/v6"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"go.expect.digital/translate/pkg/model"
+	"golang.org/x/exp/maps"
+)
+
+func mergeSlices[T ~[]E, E any](a, b T) T {
```
janishorsts, line 13: Why not simply `append(a, b...)`? If there is enough capacity, no malloc is done.

pkg/translate/field_mask_test.go
```diff
@@ -0,0 +1,282 @@
...
+
+				result.C.H = dst.C.H
+				assert.Equal(t, dst, result)
+			},
+		},
+		{
+			// Update map of strings in a nested field of struct. (Merge two maps)
+			name: "Update J.K struct.map[string]string",
+			mask: []string{"J.K"},
+			assertFunc: func(t *testing.T, dst, src, result nestedStruct) {
```
janishorsts, line 158: Why are you testing the test function - mergeMaps?

pkg/translate/field_mask_test.go
VladislavsPerkanuks: Potential side effects, I wanted to be 100% sure that merging slices and maps is pure, e.g. returns merged without modifying a or b
> I would like to emphasize @icza answer and simplify it a bit since it is a crucial concept. I assume that reader is familiar with slices.
c := append(a, b...)
This is a valid answer to the question. BUT if you need to use slices 'a' and 'c' later in code in different context, this is not the safe way to concatenate slices.

https://stackoverflow.com/a/58726780/19159194

pkg/translate/field_mask_test.go
```diff
@@ -0,0 +1,282 @@
...
+
+func mergeSlices[T ~[]E, E any](a, b T) T {
+	merged := make(T, len(a)+len(b))
+	copy(merged, a)
+	copy(merged[len(a):], b)
+
+	return merged
+}
+
+func mergeMaps[T ~map[K]V, K comparable, V any](a, b T) T {
```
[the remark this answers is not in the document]
VladislavsPerkanuks: This function uses maps.Copy to merge two maps. It was created for convenience and its name clearly states its purpose. To avoid modifying the original maps a or b, the function creates a new variable. This is because maps.Copy(a,b) appends the keys/values of b to a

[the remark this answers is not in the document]
VladislavsPerkanuks: but not relevant anymore

pkg/server/field_mask.go
```diff
@@ -22,3 +24,84 @@ func parseFieldMask(message proto.Message, paths []string) (model.Mask, error) {
...
+		tag := dst.Type().Field(i).Tag.Get("protoName")
+
+		if field != tag {
+			continue
+		}
+
+		srcField, dstField := src.Field(i), dst.Field(i)
+
+		switch dst.Field(i).Kind() { //nolint:exhaustive
+		case reflect.Struct:
```
klavinsjanis, line 69: How would protobuf types be handled? Like "google.type.Date", "google.type.DateTime" etc?

pkg/server/field_mask.go
```diff
@@ -22,3 +24,84 @@ func parseFieldMask(message proto.Message, paths []string) (model.Mask, error) {
...
+// updateFromMask updates the dst with the values from src based on the mask.
+//
+// Following scenarios are possible:
+//   - mask is nil: All fields are updated.
+//   - mask is not nil, but has no paths: No fields are updated.
+//   - mask contains paths, that does not exist in the model: The paths are ignored.
+//   - mask contains paths, that exist in model: Only those fields are updated.
+//
+// `protoName` tags are used to match fields from the fieldMask to fields in the model.
+func updateFromMask[T any](src, dst *T, mask model.Mask) *T {
```
klavinsjanis, line 37: Could use an example in the function description.

pkg/server/field_mask.go
VladislavsPerkanuks: it not supposed to update proto types.

We update our types e.g. Translation, Service then convert it to proto type later

pkg/server/field_mask.go
```diff
@@ -22,3 +24,84 @@ func parseFieldMask(message proto.Message, paths []string) (model.Mask, error) {
...
+		fields := strings.Split(path, ".")
+		updateField(reflect.ValueOf(src).Elem(), reflect.ValueOf(dst).Elem(), fields)
+	}
+
+	return dst
+}
+
+// updateField updates the dst value with the values from src, based on the fields slice.
+func updateField(src, dst reflect.Value, fields []string) {
+	if len(fields) == 0 {
```
klavinsjanis, line 53: What happens if fields contain wildcard character '*' ?

VladislavsPerkanuks: if i remember correctly 
"google.golang.org/protobuf/types/known/fieldmaskpb"

does not support `*`

pkg/server/field_mask.go
klavinsjanis: Okay, I mentioned that because there might be some struct types in the future where specific fields might not be able to be updated, only the entire struct.

pkg/server/field_mask_test.go
```diff
@@ -0,0 +1,280 @@
...
+				for _, dstElem := range dst.C.H {
+					require.Contains(t, result.C.H, dstElem)
+				}
+
+				result.C.H = dst.C.H
+				assert.Equal(t, dst, result)
+			},
+		},
+		{
+			// Update map of strings in a nested field of struct. (Merge two maps)
```
klavinsjanis, line 148: You have 'merge two maps' test case. 
Could you add a test case where a value for a specific key in the map is updated?

pkg/server/field_mask.go
VladislavsPerkanuks: then fieldMask should be according to that :)

```go
model.Mask{"foo"}
// instead of
model.Mask{"foo.bar"}
```

pkg/server/field_mask_test.go
VladislavsPerkanuks: Need implementation beforehand. For now can be ignored, as we don't have structs with maps

klavinsjanis: hmm, we could return codes.Unimplemented.

VladislavsPerkanuks: or we could ignore as we don't have maps in our updatable structs :)

klavinsjanis: Well you already added case reflect.Map :)

pkg/server/field_mask_test.go
```diff
@@ -0,0 +1,294 @@
+package server
```
klavinsjanis, line 1: Since this PR focuses on updating a service using mask, would be good to see some integration tests for the GRPC service.

VladislavsPerkanuks: They already exist, both in gRPC and REST

pkg/server/field_mask.go
```diff
@@ -22,3 +24,97 @@ func parseFieldMask(message proto.Message, paths []string) (model.Mask, error) {
...
+			}
+
+		default:
+			// For all other field kinds, set value of field in dst to value of corresponding field in src
+			dstField.Set(srcField)
+		}
+	}
+}
+
+// updateServiceFromMask updates the dstService with the values from srcService based on the fieldMask.
```
janishorsts, line 111: If we apply zero tolerance to a modification of read-only values, we can define the functionality around it.

If somebody attempts to update a read-only field like `id`, a service rejects the request. Bad request.

Any ideas?

janishorsts (commented):
My concern is transport serialization layer has leaked into models.

pkg/server/field_mask.go
VladislavsPerkanuks: for now, this should suffice.
Invalid argument error on trying to update service.ID

pkg/server/field_mask.go
```diff
@@ -1,24 +1,136 @@
...
+		return
+	}
+}
+
+// updateServiceFromMask updates the dstService with the values from srcService based on the fieldMask.
+func updateServiceFromMask(
+	srcService *model.Service,
+	dstService *model.Service,
+	mask model.Mask,
+) error {
```
janishorsts, line 124:
```suggestion
func updateServiceFromMask(src, dst *model.Service, mask model.Mask) error {
```

pkg/server/transform.go
```diff
@@ -212,3 +217,72 @@ func translationFromProto(t *translatev1.Translation) (*model.Translation, error
...
+
+func maskFromProto(message proto.Message, mask *fieldmaskpb.FieldMask) (model.Mask, error) {
+	if mask == nil {
+		return nil, nil
+	}
+
+	paths := mask.GetPaths()
+
+	if len(paths) == 0 {
+		return model.Mask{}, nil
```
janishorsts, line 231: Always return nil slice. Mostly, an empty slice is required only when marshalling JSON.

```suggestion
		return nil, nil
```

VladislavsPerkanuks: it changes functionality:` nil mask != empty mask`

nil mask means update dst with all values from src
empty mask means update nothing, e.g. leave dst as is

VladislavsPerkanuks: nil, is when in request fieldMask is not specified, eg left out
empty is when it is specified in the request, but left empty

pkg/server/transform.go
```diff
@@ -212,3 +217,72 @@ func translationFromProto(t *translatev1.Translation) (*model.Translation, error
...
+	}
+
+	// Find corresponding model type to proto message.
+	var modelType any
```
janishorsts, line 235: You should be able to use protoreflect and avoid explicit type implementation. Right? It is less expensive too.

```go
m.Protoreflect().Descriptor()
```

pkg/server/transform.go
janishorsts: looks confusing to me. why would anybody want to update without updating? empty (nil) means implicit full update. nice and clear. what do you think?

VladislavsPerkanuks: >  If a FieldMask object is not present in a get operation, the
 operation applies to all fields (as if a FieldMask of all fields
 had been specified).

https://protobuf.dev/reference/csharp/api-docs/class/google/protobuf/well-known-types/field-mask

Same goes for update operations.

in the document, there is no information on how API should behave when an empty fieldmask is received, so I infer that an empty mask is a valid mask, and updates nothing.

> . The API is required to only change the values of the fields as specified in the mask and leave the others untouched.

no paths == leave everything untouched

Of course, we could change the implementation, so if field mask is provided it must contain at least one path, else throw an error, but by doing so we are violating design...

VladislavsPerkanuks: Answering, on this
> why would anybody want to update without updating?

why would one provide fieldmask field in request, and leave it empty. You either provide it and fill paths you want to update, or don't provide it at update everything 🤷

janishorsts: The intent is the same whether the field mask was set empty or not set at all, isn't it?

pkg/server/transform.go
VladislavsPerkanuks: okay, i understand what you meant, but still wont work as some fields differ 

proto generated Message, has `PluralId` when we in model we have `PluralID`...
image

VladislavsPerkanuks: same for `Id` In the future those differences will only increase in size

pkg/server/transform.go
VladislavsPerkanuks: who knows the one's intent in this scenario :)

pkg/server/transform.go
janishorsts: The names are identical, but the case may differ.

pkg/server/transform.go
janishorsts: Simplicity is your sherpa :)

pkg/server/transform.go
VladislavsPerkanuks: > Make it case-insensitive?

I did not think of that. That actually might and should work, and that would simplify everything a lot 👍

VladislavsPerkanuks: Resolving edge cases is a big no-no. It is now hard to work on the project (somehow), and adding these workarounds will only make situation worse

pkg/server/transform.go
VladislavsPerkanuks: I would go with the error option then
if fieldmask not provided -> update all 
if fieldmask provided and empty -> error fieldmask should contain at least one path

It seems the most correct/logical one, at least for me

janishorsts (approved):

PR #71, opened 2023-05-03
Build Translate Image

VladislavsPerkanuks:
## Changes:
- Added missing env variables to `.env.test`
- Added reflection to gRPC server to assist clients
- Improved `+deps` and `+lint-proto` targets
- Added `+build`, `+image` and `image-multiplatform`, to compile go code to binary, create image, and create multiplatform image respectively 
- Bumped dependencies versions
- Removed redundant print statements in `parse_test`


Earthfile
```diff
@@ -2,16 +2,16 @@ VERSION 0.7
...
-ARG --global sqlfluff_version=2.0.7
+ARG --global sqlfluff_version=2.1.0
 
 FROM golang:$go_version-alpine
 
 deps:
   WORKDIR /translate
   COPY go.mod go.sum .
-  RUN go mod download
+  RUN --mount=type=cache,target=/go/pkg/mod go mod download
```
janishorsts, line 14: When you use a cache mount, it is constrained to only the RUN task you run. `+go` does not benefit from it - it downloads all dependencies again.

The frequency of changes in go.mod is rare, it's more beneficial to not use a cached mount.

Earthfile
```diff
@@ -130,3 +128,30 @@ test-integration:
...
+
+# -----------------------Building-----------------------
+
+build:
+  COPY +go/translate translate
+  WORKDIR translate
+  RUN \
+  --mount=type=cache,target=/go/pkg/mod \
+  --mount=type=cache,target=/root/.cache/go-build \
+    CGO_ENABLED=0 go build -o translate cmd/translate/main.go
```
janishorsts, line 140: It keeps it cleaner to set env variables using ENV.

```suggestion
  ENV CGO_ENABLED=0 
  RUN \
  --mount=type=cache,target=/go/pkg/mod \
  --mount=type=cache,target=/root/.cache/go-build \
     go build -o translate cmd/translate/main.go
```

Earthfile
```diff
@@ -130,3 +130,34 @@ test-integration:
 test:
   BUILD +test-unit
   BUILD +test-integration
+
+# -----------------------Building-----------------------
+
+build:
+  ARG GOARCH=arm64
+  ENV CGO_ENABLED=0 
+  ENV GOARCH=$GOARCH
```
janishorsts, line 139: All ARGs are available as ENVs in the container, we can safely remove ENV GOARCH.

```suggestion
```

Earthfile
```diff
@@ -130,3 +130,34 @@ test-integration:
 test:
   BUILD +test-unit
   BUILD +test-integration
+
+# -----------------------Building-----------------------
+
+build:
+  ARG GOARCH=arm64
```
janishorsts, line 137: Does this work?

```suggestion
  ARG GOARCH=$USERARCH
```

Earthfile
VladislavsPerkanuks: yeah, I was curios about that, could not find information

Earthfile
VladislavsPerkanuks: yes. missed this

Earthfile
```diff
@@ -130,3 +130,34 @@ test-integration:
...
+  COPY --platform=linux/$USERARCH (+build/bin/translate --GOARCH=$TARGETARCH) /translate
+  ENTRYPOINT ["/translate"]
+  SAVE IMAGE --push $registry/translate:$tag
+
+image-multiplatform:
+  ARG --required registry
+  BUILD \
+  --platform=linux/amd64 \
+  --platform=linux/arm/v7 \
+  --platform=linux/arm64 \
```
janishorsts, line 162: Let's support the following initially!

```suggestion
  --platform=linux/amd64 \
  --platform=linux/arm64 \
```

janishorsts (approved):

PR #74, opened 2023-05-15
48-add-translate-cli

klavinsjanis:


cmd/client/cmd/download.go
```diff
@@ -0,0 +1,127 @@
...
+				fileName += "." + arb
+			case translatev1.Schema_JSON_NG_LOCALIZE, translatev1.Schema_JSON_NGX_TRANSLATE, translatev1.Schema_GO:
+				fileName += "." + json
+			case translatev1.Schema_POT:
+				fileName += "." + pot
+			case translatev1.Schema_XLIFF_12, translatev1.Schema_XLIFF_2:
+				fileName += "." + xlf
+			}
+
+			if err = os.WriteFile(filepath.Join(path, fileName), res.Data, 0644); err != nil { //nolint:gomnd,gofumpt,gosec
```
klavinsjanis, line 87: Not quite sure about setting file permissions, maybe we could pass in the permissions as a cli parameter?

DarjaLeonova (approved):

cmd/client/cmd/list.go
```diff
@@ -0,0 +1,53 @@
...
+		RunE: func(cmd *cobra.Command, args []string) error {
+			timeout, err := cmd.InheritedFlags().GetDuration("timeout")
+			if err != nil {
+				return fmt.Errorf("list services: get cli parameter 'timeout': %w", err)
+			}
+
+			ctx, cancelFunc := context.WithTimeout(context.Background(), timeout)
+			defer cancelFunc()
+
+			client, err := newClientConn(ctx, cmd)
```
VladislavsPerkanuks, line 26: instead of creating client for every operation, how about using package level variable to declare only once ?

cmd/client/cmd/root.go
```diff
@@ -0,0 +1,116 @@
...
+func (s *schema) Set(v string) error {
+	switch v {
+	case "json_ng_localize", "json_ngx_translate", "go", "arb", "pot", "xliff_12", "xliff_2":
+		*s = schema(v)
+		return nil
+	default:
+		return errors.New(
+			`must be one of "json_ng_localize", "json_ngx_translate", "go", "arb", "pot", "xliff_12", "xliff_2"`)
+	}
+}
```
VladislavsPerkanuks, line 103:
```suggestion
func (s *schema) Set(v string) error {
	if _, ok := translatev1.Schema_value[strings.ToUpper(v)]; ok {
		*s = schema(v)
		return nil
	}

	validSchemas := make([]string, 0, len(translatev1.Schema_name))
	for _, name := range translatev1.Schema_name {
		validSchemas = append(validSchemas, strings.ToLower(name))
	}

	return fmt.Errorf("Invalid schema value. Must be one of: %s", strings.Join(validSchemas, ", "))
}
```

1. all possible schemas are already definied in translatev1
2. Error message is constructed dynamicly

cmd/translate/main_client_integration_test.go
```diff
@@ -0,0 +1,314 @@
...
+			"messages":[
+				 {
+					"id":"1",
+					"meaning":"When you greet someone",
+					"message":"hello",
+					"translation":"čau",
+					"fuzzy":false
+				 }
+			]
+	 }`))
```
VladislavsPerkanuks, line 66: You could use here `randUploadData()` which is definied in `grpc_integration_test.go`, to avoid defining upload data all over again
There are similar functions as well, you could check it out, so tests will become cleaner

cmd/translate/main_client_integration_test.go
```diff
@@ -0,0 +1,314 @@
...
+
+		require.NoError(t, err)
+
+		res, err := cmd.ExecuteWithParams([]string{
+			"service", "upload",
+			"-a", fmt.Sprintf("%s:%s", host, port),
```
VladislavsPerkanuks, line 72: I think package level variable addr would be more appropriate here :)
```
var addr = host + ":" + port
```

cmd/translate/main_client_integration_test.go
```diff
@@ -0,0 +1,314 @@
...
+	})
+}
+
+func Test_TranslationFileUpload(t *testing.T) {
+	t.Run("OK", func(t *testing.T) {
+		service, err := client.CreateService(context.Background(),
+			&translatev1.CreateServiceRequest{Service: randService()})
+
+		require.NoError(t, err)
+		require.NotNil(t, service)
```
VladislavsPerkanuks, line 49:
```suggestion
    service := createService(context.Background(), t *testing.T)
```

Same funcitionality already defined in grpc_integration_test

cmd/client/cmd/list.go
klavinsjanis: A new client will be created for each command call, and the client will be set up using the provided cmd flags.

cmd/client/cmd/root.go
```diff
@@ -0,0 +1,121 @@
...
+		return nil
+	}
+
+	validSchemas := make([]string, len(translatev1.Schema_name))
+	for schemaValue, schemaName := range translatev1.Schema_name {
+		validSchemas[schemaValue] = strings.ToLower(schemaName)
+	}
+
+	return fmt.Errorf("invalid schema value: must be one of: %s", strings.Join(validSchemas[1:], ", "))
+}
```
VladislavsPerkanuks, line 108: how about creating validSchema slice in `init()`, so the slice is created only once ?
```suggestion
var validSchemas []string
func init() {
	validSchemas = make([]string, len(translatev1.Schema_name))
	for schemaValue, schemaName := range translatev1.Schema_name {
		validSchemas[schemaValue] = strings.ToLower(schemaName)
	}
}

// Set must have pointer receiver so it doesn't change the value of a copy.
func (s *schema) Set(v string) error {
	v = strings.ToUpper(v)

	if sv, ok := translatev1.Schema_value[v]; ok && sv != 0 {
		*s = schema(v)
		return nil
	}

	return fmt.Errorf("invalid schema value: must be one of: %s", strings.Join(validSchemas[1:], ", "))
}
```

VladislavsPerkanuks (approved):

cmd/client/cmd/root.go
VladislavsPerkanuks: and even better to not include UNSPECIFIED schema to validSchemas :)
e.g.
```go
validSchemas = make([]string, len(translatev1.Schema_name)-1) // all schemas, exluding unspecified
	for schemaValue, schemaName := range translatev1.Schema_name {
                if schemaValue != 0 {
		  validSchemas[schemaValue-1] = strings.ToLower(schemaName)
                 }
	}
...

	return fmt.Errorf("invalid schema value: must be one of: %s", strings.Join(validSchemas, ", "))

```

janishorsts (changes requested):
* please move CLI integration tests to `cmd/client`
* please remove all aliases for commands
* please ensure CLI integration tests send traces

janishorsts (approved):
I did not review the code - I trust you @klavinsjanis. We'll deal with any technical debt later.

PR #78, opened 2023-06-08
Add Google Translate API 

VladislavsPerkanuks:
# Changes
-  Renamed existing `translate` pkg to `server`  
-  Added new package `translate`, with sub-package `googletranslate`
-  Added Interface with `Translate()` func 
```go
type TranslationService interface {
	Translate(ctx context.Context, message *model.Messages, targetLang language.Tag) (*model.Messages, error)
}
```
- Added Google Translate implementation
- Added unit tests for Google Translate `Translate()` method, using mocks
- Added integration tests for Google Translate `Translate()` method using real API (skip tests if API key not defined)


pkg/server/server.go
VladislavsPerkanuks, line 1: Same as before, just renamed from `translate.go` to `server.go`

.env.test
```diff
@@ -5,15 +5,18 @@ OTEL_RESOURCE_ATTRIBUTES=deployment.environment=test
...
 TRANSLATE_DB_MYSQL_PORT=3306
 TRANSLATE_DB_MYSQL_USER=root
 TRANSLATE_DB_MYSQL_PASSWORD=
 TRANSLATE_DB_MYSQL_DATABASE=translate
 
 # BadgerDB
 TRANSLATE_DB_BADGERDB_PATH=/tmp/badger
+
+# Google Translate API
+TRANSLATE_TRANSLATE_SERVICES_GOOGLE_TRANSLATE_API_KEY=
```
klavinsjanis, line 22: seems a bit too verbose. Word TRANSLATE gets repeated 3 times.

pkg/server/convert.go
```diff
@@ -1,4 +1,4 @@
-package translate
+package server
```
klavinsjanis, line 1: package server? didn't you mean service?

pkg/server/server.go
```diff
@@ -0,0 +1,15 @@
+package server
+
+import (
+	translatev1 "go.expect.digital/translate/pkg/pb/translate/v1"
+	"go.expect.digital/translate/pkg/repo"
+)
+
+type TranslateServiceServer struct {
```
klavinsjanis, line 8: Why not just TranslateService?

pkg/translate/translatetest/util.go
```diff
@@ -0,0 +1,16 @@
+package translatetest
```
klavinsjanis, line 1: why not keep this in pkg/testutil/rand folder?

.env.test
VladislavsPerkanuks: first translate is a prefix, not debatble

second translate is the one which could be debatable

third one is a google_translate, not debatble

pkg/server/server.go
VladislavsPerkanuks: because it has, so logicly this one is ImplementedTranslateServiceServer or shorter `TranslateServiceServer`
`translatev1.UnimplementedTranslateServiceServer`

pkg/translate/translatetest/util.go
VladislavsPerkanuks: it is specificly tailored for 3rd party translate services integration tests, so i thought it is logicly that is in the same pkg

pkg/server/convert.go
VladislavsPerkanuks: this one debatable, maybe you right, @janishorsts what you think, service or server ?

.env.test
klavinsjanis: how about GOOGLE_TRANSLATE_API_KEY ?

VladislavsPerkanuks: the `TRANSLATE` prefix must be present
and we will have multiple translate_services, so some intermediate is needed, and `translate_services` should suffice
```yaml
translate_services:
  google_translate:
    api_key: ...
    #others service specific keys
  another_service:
   ...
```

this is `examples/translate.yaml` and env variables must be the same layout, e.g. prefix + path to key, and in this case `TRANSLATE_TRANSLATE_SERVICES_GOOGLE_TRANSLATE_API_KEY`

Earthfile
```diff
@@ -105,7 +105,7 @@ test-integration:
   COPY +go/translate /translate
   COPY --dir migrate/mysql migrate
   WITH DOCKER --compose compose.yaml --service mysql --pull migrate/migrate:v$migrate_version --pull golang:$go_version-alpine
-    RUN \
+    RUN --secret=googletranslate_api_key \
```
janishorsts, line 108: TRANSLATE_TRANSLATE_SERVICES_GOOGLE_TRANSLATE_API_KEY is unsafely passed to docker run as an environment variable. I don't know what the implications are but let's be on the safe side. I hope it is not saved in the BuildKit layers.

Use `RUN --no-cache`.

janishorsts: It would be nice to find out from earthly

.env.test
janishorsts: We have to simplify it. I suggest a general purpose `TRANSLATE_OTHER` settings for external APIs - TRANSLATE_OTHER_GOOGLE_TRANSLATE_API_KEY.

pkg/server/convert.go
janishorsts: keep `server`

pkg/translate/translate.go
janishorsts, line 1: We can improve the name of this package. Its current name might be confusing in the context of the project.

This may help us in the future if we start adding package descriptions.

This package and sub-packages implement fuzzy translations from external ML/translation services.

`package fuzzy`?

janishorsts, line 1: This package/sub-packages are very simple. Let's bring them all under a single package `fuzzy`.

Earthfile
VladislavsPerkanuks: Tried to find something about that, but no luck with that. For now, will stick with `--no-cache` (in **ci/cd** pipeline it does not matter anyways as every run is run in sandbox environment, this will only affect local integration test execution speed, when running `earthly +test-integration` )

pkg/fuzzy/fuzzy.go
```diff
@@ -0,0 +1,15 @@
...
+import (
+	"context"
+
+	"go.expect.digital/translate/pkg/model"
+	"golang.org/x/text/language"
+)
+
+var SupportedServices = []string{"GoogleTranslate"}
+
+type TranslationService interface {
```
janishorsts, line 12:
```suggestion
type Translator interface {
```

pkg/fuzzy/googletranslate.go
```diff
@@ -0,0 +1,137 @@
...
+// Interface that defines some of the methods of the Google Translate client.
+// This interface helps to mock the Google Translate client in unit tests.
+// https://pkg.go.dev/cloud.google.com/go/translate#Client
+type googleClient interface {
+	Translate(ctx context.Context, inputs []string, target language.Tag, opts *translate.Options) ([]translate.Translation, error) //nolint:lll
+	io.Closer
+}
+
+// googleTranslate implements the TranslationService interface.
+type googleTranslate struct {
```
janishorsts, line 29: NewGoogleTranslate returns this struct - always return an exposed identifier in the exposed function.

```suggestion
type GoogleTranslate struct {
```

pkg/fuzzy/googletranslate.go
```diff
@@ -0,0 +1,137 @@
...
+func (g *googleTranslate) Translate(
+	ctx context.Context,
+	messages *model.Messages,
+	targetLang language.Tag,
+) (*model.Messages, error) {
+	if err := validateTranslate(messages, targetLang); err != nil {
+		return nil, fmt.Errorf("google translate: validate translate request: %w", err)
+	}
+
+	translatedMessages := model.Messages{
```
janishorsts, line 105: Move this down closer to consumption! Always make sure it is efficient!

You declare a variable and reserve memory here, but if the translation fails, it was a wasted effort.

pkg/fuzzy/validate.go
```diff
@@ -0,0 +1,34 @@
...
+	errNilMessages         error = errors.New("nil messages")
+	errNoMessages          error = errors.New("no messages")
+	errTargetLangUndefined error = errors.New("target language undefined")
+)
+
+// validateTranslate validates the given messages and target language
+// before executing translation request to save time and resources.
+//
+// TODO: Add support for validating for source and target language support by the translation service.
+func validateTranslate(messages *model.Messages, targetLang language.Tag) error {
```
janishorsts, line 20: We don't require data validation we trust. We always trust the data inside the system.

pkg/fuzzy/googletranslate.go
```diff
@@ -0,0 +1,137 @@
...
+// --------------------Methods--------------------
+
+func (g *googleTranslate) Translate(
+	ctx context.Context,
+	messages *model.Messages,
+	targetLang language.Tag,
+) (*model.Messages, error) {
+	if err := validateTranslate(messages, targetLang); err != nil {
+		return nil, fmt.Errorf("google translate: validate translate request: %w", err)
+	}
```
janishorsts, line 103: simply do

```suggestion
	if messages == nil {
	  return nil, nil
	}
	
	if len(messages.Messages) == 0 {
	  return &model.Messages{Language: targetLang}
	}
```

pkg/fuzzy/googletranslate.go
janishorsts, line 1: google_translate.go

pkg/fuzzy/googletranslate.go
```diff
@@ -0,0 +1,142 @@
...
+	htransport "google.golang.org/api/transport/http"
+)
+
+// --------------------Definitions--------------------
+
+// Interface that defines some of the methods of the Google Translate client.
+// This interface helps to mock the Google Translate client in unit tests.
+// https://pkg.go.dev/cloud.google.com/go/translate#Client
+type googleClient interface {
+	Translate(ctx context.Context, inputs []string, target language.Tag, opts *translate.Options) ([]translate.Translation, error) //nolint:lll
```
janishorsts, line 24: alternatively, the types are self explanatory

```suggestion
	Translate(context.Context, []string, language.Tag, *translate.Options) ([]translate.Translation, error)
```

janishorsts (approved):

PR #82, opened 2023-06-12
Add Support For BadgerDB

VladislavsPerkanuks:
# Changes:
- Added BadgerDB support (Repo interface implementation and Repo initialization)
- Moved integration tests, to `pkg/repo/repotest`, so all integrations could share tests
- Changed default DB to BadgerDB, and added support for multiple DB's
- Added new pkg, `testutil/rand`, for unified random data generation for common types across all tests, e.g. Messages, message, service


go.mod
```diff
@@ -28,27 +28,40 @@ require (
 
 require (
 	github.com/cenkalti/backoff/v4 v4.2.1 // indirect
+	github.com/cespare/xxhash v1.1.0 // indirect
+	github.com/cespare/xxhash/v2 v2.2.0 // indirect
 	github.com/davecgh/go-spew v1.1.1 // indirect
+	github.com/dgraph-io/badger/v3 v3.2103.5
```
klavinsjanis, line 34: shouldn't this be in the direct imports?

VladislavsPerkanuks: yes, missed that one

pkg/repo/badgerdb/message.go
```diff
@@ -0,0 +1,81 @@
...
+		return &messages, nil // Empty messages.messages for this service (Not an error)
+	case err != nil:
+		return nil, fmt.Errorf("repo: load service: %w", err)
+	}
+
+	err = r.db.View(func(txn *badger.Txn) error {
+		item, getErr := txn.Get(getMessagesKey(serviceID, language))
+		switch {
+		default:
+			// noop
```
klavinsjanis, line 67: isn't the default behaviour to call getValues(item, &messages) function ?

pkg/repo/badgerdb/message.go
```diff
@@ -0,0 +1,81 @@
...
+	return nil
+}
+
+func (r *Repo) LoadMessages(ctx context.Context, serviceID uuid.UUID, language language.Tag) (*model.Messages, error) {
+	messages := model.Messages{Language: language}
+
+	_, err := r.LoadService(ctx, serviceID)
+
+	switch {
+	default:
```
klavinsjanis, line 55: default not needed

VladislavsPerkanuks: default is not needed, you are right, but I choose to keep it as it makes this switch statement more explicit, e.g. explicitly tell that when everything is okay (default case), nothing should be done, instead of implicitly telling by avoiding default case

pkg/repo/badgerdb/message.go
VladislavsPerkanuks: yes, ty

pkg/testutil/rand/model.go
```diff
@@ -0,0 +1,160 @@
...
+	count uint,
+	diffLang bool,
+	msgOpts []ModelMessageOption,
+	opts ...ModelMessagesOption,
+) []*model.Messages {
+	modelMsg := func(opts ...ModelMessagesOption) *model.Messages {
+		return ModelMessages(3, msgOpts, opts...) //nolint:gomnd
+	}
+
+	msgsSlice := slice(count, modelMsg, opts...)
```
klavinsjanis, line 137: msgsSlice, why not just messages?

VladislavsPerkanuks: because messages is usually model.Messages
in this case it is []model.Messages

pkg/repo/common/error.go
VladislavsPerkanuks, line 1: circular import error workaround

pkg/repo/create.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+func (r RepoDB) String() string {
+	return RepoDBNames[r]
+}
+
+// SupportedDBs returns a list of supported databases.
+func SupportedDBs() []string {
+	dbs := make([]string, 0, len(RepoDBNames))
+	for _, v := range RepoDBNames {
```
klavinsjanis, line 31: Just a note: when ranging over a map element order can change.

pkg/translate/service.go
```diff
@@ -8,7 +8,7 @@ import (
 	"github.com/google/uuid"
 	"go.expect.digital/translate/pkg/model"
 	translatev1 "go.expect.digital/translate/pkg/pb/translate/v1"
-	"go.expect.digital/translate/pkg/repo"
+	"go.expect.digital/translate/pkg/repo/common"
```
janishorsts, line 11: why change from `repo` to `common`?

VladislavsPerkanuks: https://github.com/translate-agent/translate/pull/82#discussion_r1235162327

I wanted to shorten/lighten the main function of creating a server, by moving the newRepo creation to repo pkg. 

as every implementation imports repo pkg (for now only for ErrNotFound), and repo pkg imports implementation-specific NewRepo(), circular import is encountered

pkg/repo/badgerdb/repo.go
```diff
@@ -0,0 +1,65 @@
...
+		if path == "" {
+			return fmt.Errorf("WithDefaultDB: viper: empty db.badgerdb.path")
+		}
+
+		var err error
+		r.db, err = newDB(badger.DefaultOptions(path))
+
+		if err != nil {
+			return fmt.Errorf("WithDefaultDB: open badger db: %w", err)
+		}
```
janishorsts, line 31: keeps err handling closer

```suggestion
		var err error
		
		r.db, err = newDB(badger.DefaultOptions(path))
		if err != nil {
			return fmt.Errorf("WithDefaultDB: open badger db: %w", err)
		}
```

pkg/repo/badgerdb/service.go
```diff
@@ -0,0 +1,132 @@
...
+)
+
+const servicePrefix = "service:"
+
+// getServiceKey converts a service ID to a BadgerDB key with prefix.
+func getServiceKey(id uuid.UUID) []byte {
+	return []byte(fmt.Sprintf("%s%s", servicePrefix, id))
+}
+
+// getValue unmarshals the value of a BadgerDB item into v (either *model.Service or *model.Messages).
```
janishorsts, line 22: Move this function to `badgersdb.go` so we have common functionality of the package in one place.

pkg/testutil/rand/model.go
```diff
@@ -0,0 +1,155 @@
...
+	"go.expect.digital/translate/pkg/model"
+	"golang.org/x/text/language"
+)
+
+type modelType interface {
+	model.Service | model.Message | model.Messages
+}
+
+// mdl returns a random modelType. It uses the given randF to generate the modelType.
+func mdl[T modelType, O ~func(*T)](randF func() *T, opts ...O) *T {
```
klavinsjanis, line 15: This function name doesn't convey what it does.

pkg/repo/common/error.go
janishorsts: ok, I got why `common` was created now.

pkg/testutil/rand/model.go
VladislavsPerkanuks: i was stuck with function name as model is reserved already by pkg 
if there is some better ideas, i am all ears :)

pkg/repo/create.go
```diff
@@ -0,0 +1,76 @@
...
+		dbs = append(dbs, v)
+	}
+
+	return dbs
+}
```
janishorsts: Why not simply have a list of supported databases? It simplifies the whole thing.

```suggestion
var supportedDBs = [...]string{"mysql", "badgerdb"}
```

VladislavsPerkanuks: i guess i had something in mind yesterday, but yeah looking now it looks kinda pointless

pkg/testutil/rand/model.go
klavinsjanis: mdl -> randomModelType ?
slice -> randomModelTypes ?

pkg/repo/repotest/integration_test.go
```diff
@@ -0,0 +1,85 @@
...
+	if err := initBadgerDB(); err != nil {
+		log.Println(err)
+		os.Exit(1)
+	}
+
+	os.Exit(m.Run())
+}
+
+// allRepos runs a test for each repo that is defined in the repos map.
+func allRepos(t *testing.T, f func(t *testing.T, repo repo.Repo, subtest testutil.SubtestFn)) {
```
klavinsjanis, line 74: alternative: testAllRepos

pkg/testutil/rand/model.go
```diff
@@ -0,0 +1,155 @@
...
+		s = append(s, f(opts...))
+	}
+
+	return s
+}
+
+// ------------------Service------------------
+
+// modelService generates a random model.Service.
+func modelService() *model.Service {
```
klavinsjanis, line 38: alternative: randomModelService

VladislavsPerkanuks: pkg name is rand, it implies it already

pkg/testutil/rand/model.go
VladislavsPerkanuks: i need to think about that

pkg/repo/repotest/integration_test.go
VladislavsPerkanuks: allRepos seems better, but dunno

janishorsts (approved):

PR #85, opened 2023-06-14
Add CORS

VladislavsPerkanuks:


VladislavsPerkanuks:
In this situation, adding CORS would not be effective because grpc-web, which is used by all browsers for requests, is different from grpc. A “translator” that converts between grpc-web and grpc is necessary for successful client-to-server requests. Using Envoy as a proxy can solve this issue and it can also be configured with CORS policies, making this PR unnecessary.

PR #96, opened 2023-06-22
94-list-messages

klavinsjanis:


pkg/translate/transform.go
```diff
@@ -53,6 +53,22 @@ func langTagFromProto(s string) (language.Tag, error) {
 	return l, nil
 }
 
+// langTagsFromProto converts a slice of strings to []language.Tag.
+func langTagsFromProto(langTags []string) ([]language.Tag, error) {
```
VladislavsPerkanuks, line 57: i guess you could change generic function, so if input is pointer then output is also list with pointers, and vice versa
```go
// sliceFromProto converts a slice of type T to a slice of type R using the provided elementFromProto function.
func sliceFromProto[T any, R any](slice []T, elementFromProto func(T) (R, error)) ([]R, error) {
...
}

// langTagsFromProto converts a slice of strings to []language.Tag.
func langTagsFromProto(langTags []string) ([]language.Tag, error) {
	return sliceFromProto(langTags, langTagFromProto)
}
```

pkg/repo/repotest/message_integration_test.go
```diff
@@ -171,38 +175,98 @@ func Test_LoadMessages(t *testing.T) {
...
+		langTags = filter.LanguageTags(langTags)
+		messages := make([]model.Messages, 0, len(langTags))
+
+		for _, langTag := range langTags {
+			msgs := rand.ModelMessages(1, nil, rand.WithLanguage(langTag))
+			err := repository.SaveMessages(testCtx, service.ID, msgs)
+			require.NoError(t, err, "Prepare test messages")
+
+			messages = append(messages, *msgs)
+		}
```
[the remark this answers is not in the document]
klavinsjanis: umm, doesn't look like this reduces the boilerplate code by much.

pkg/repo/mysql/message.go
```diff
@@ -99,36 +101,70 @@ ON DUPLICATE KEY UPDATE
...
-		`SELECT mm.id, mm.message, mm.description, mm.fuzzy
-FROM message_message mm
-JOIN message m ON m.id = mm.message_id
-WHERE m.service_id = UUID_TO_BIN(?) AND m.language = ?`,
-		serviceID,
-		language.String(),
+func (r *Repo) LoadMessages(ctx context.Context, serviceID uuid.UUID, opts common.LoadMessagesOpts,
+) ([]model.Messages, error) {
+	var (
+		qb   strings.Builder
```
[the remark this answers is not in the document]
klavinsjanis: Well, there are many options, I guess we have to decide on how we are going to build SQL queries in this project.

pkg/translate/transform.go
klavinsjanis: I think I might have to create another generic function, function signatures are a bit different for service transformations.

pkg/translate/message.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+
+	return &listMessagesParams{serviceID: serviceID, languageTags: langTags}, nil
+}
+
+func (l *listMessagesParams) validate() error {
+	if l.serviceID == uuid.Nil {
+		return errors.New("'service_id' is required")
+	}
+
```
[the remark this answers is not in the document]
klavinsjanis: umm, language is not required.

proto/translate/v1/translate.proto
```diff
@@ -53,6 +53,17 @@ message DownloadTranslationFileResponse {
   bytes data = 1;
 }
 
+// --------------Messages requests/responses-------------------
+
+message ListMessagesRequest {
+  string service_id = 1;
+  repeated string languages = 2;
```
janishorsts, line 60: Scope creep? Do we need `languages`? When we do filtering we'll do based on https://google.aip.dev/160.

klavinsjanis: removed languages filter

pkg/repo/badgerdb/message.go
```diff
@@ -47,34 +47,90 @@ func (r *Repo) SaveMessages(ctx context.Context, serviceID uuid.UUID, messages *
...
+				return nil // Empty messages.messages for this language (Not an error)
+			case er != nil:
+				return fmt.Errorf("transaction: get messages for language tag '%s': %w", langTag, er)
+			}
+		}
+
+		return nil
+	}); err != nil {
+		return nil, fmt.Errorf("repo: db view: %w", err)
+	}
```
VladislavsPerkanuks, line 105: two things that personally i would change, that in my opinion could improve readability
1. Using syntax `if err := ...; err != nil`, is only hurting readability when if body is long, for those scenarios standard
```go
err := ...
if err != nil {}
```
2. As there are 3 error varriables, i would give them appropriate namings
```go
viewErr := r.db.View...
getErr := txn.Get...
valErr := getValue...
```

pkg/repo/mysql/message.go
```diff
@@ -99,36 +100,58 @@ ON DUPLICATE KEY UPDATE
...
 
 	if err := rows.Err(); err != nil {
 		return nil, fmt.Errorf("repo: scan messages: %w", err)
 	}
 
+	messages := make([]model.Messages, 0, len(messagesLookup))
+
+	for langTag, msgs := range messagesLookup {
+		messages = append(messages, model.Messages{
+			Language: language.MustParse(langTag),
```
VladislavsPerkanuks, line 140: error handling instead of panic e.g. MustParse -> Parse.

for this case i think you could define struct that implements Scanner and Valuer, to get rid off manual parsing in all repo functions

e.g.

```go
type lang struct {
  language.Tag
}

func (l lang) Value() (driver.Value, error) {
...
}

func (l *lang) Scan(value interface{}) error {
...
}
```

pkg/repo/badgerdb/message.go
klavinsjanis: hmm, maybe could replace er with txErr, other stuff seems fine.

pkg/repo/mysql/message.go
janishorsts: In the end, both ways are right. The probability of it happening is very low, we always ensure we keep clean and sanitized data in our system. Users still can edit data directly in the database which happens a lot.

`driver.Valuer` and `sql.Scanner` would be nicer, the same implementation across multiple SQL repo implementations. Leaving it as tech debt.

PR #108, opened 2023-06-29
updated messages in grpc

DarjaLeonova:


cmd/translate/main_grpc_integration_test.go
```diff
@@ -419,6 +419,20 @@ func Test_ListServices_gRPC(t *testing.T) {
 
 // ------------------Messages------------------
 
+func randMessages() *translatev1.Messages {
```
klavinsjanis, line 422: I believe this already exists in the main branch, please merge main into working branch.

cmd/translate/main_grpc_integration_test.go
```diff
@@ -470,3 +484,55 @@ func Test_ListMessages_gRPC(t *testing.T) {
...
+
+	// Requests
+
+	tests := []struct {
+		request      *translatev1.UpdateMessagesRequest
+		name         string
+		expectedCode codes.Code
+	}{
+		{
+			name:         "Happy path, get all messages",
```
klavinsjanis, line 510: 'get all messages' might be misleading -> 'update successful' might be more appropriate.

cmd/translate/main_grpc_integration_test.go
```diff
@@ -470,3 +484,55 @@ func Test_ListMessages_gRPC(t *testing.T) {
...
+			request:      &translatev1.UpdateMessagesRequest{ServiceId: service.Id, Messages: randMessages()},
+			expectedCode: codes.OK,
+		},
+		{
+			name:         "Happy path, service doesn't exist",
```
klavinsjanis, line 515: this isn't happy path -> codes.NotFound

pkg/translate/message.go
```diff
@@ -55,3 +58,72 @@ func (t *TranslateServiceServer) ListMessages(
...
+	serviceID uuid.UUID
+}
+
+func parseUpdateMessagesRequestParams(req *translatev1.UpdateMessagesRequest) (*updateMessagesParams, error) {
+	serviceID, err := uuidFromProto(req.GetServiceId())
+	if err != nil {
+		return nil, fmt.Errorf("parse service_id: %w", err)
+	}
+
+	requestLanguageTag, err := language.Parse(req.GetMessages().Language)
```
klavinsjanis, line 75: I believe you should use messagesFromProto() to perform messages transformation

cmd/translate/main_grpc_integration_test.go
```diff
@@ -419,6 +419,20 @@ func Test_ListServices_gRPC(t *testing.T) {
```
klavinsjanis, line 419: I am not quite sure about the scope of this PR, but shouldn't there be tests for http requests as well?

pkg/translate/message.go
```diff
@@ -55,3 +58,72 @@ func (t *TranslateServiceServer) ListMessages(
...
+	params, err := parseUpdateMessagesRequestParams(req)
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, err.Error())
+	}
+
+	if err = params.validate(); err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, err.Error())
+	}
+
+	err = t.repo.SaveMessages(ctx, params.serviceID, params.messages)
```
klavinsjanis, line 118: check for ErrNotFound and return codes.NotFound if service is not found.

cmd/translate/main_grpc_integration_test.go
VladislavsPerkanuks: +1 Add REST integration tests

pkg/translate/message.go
VladislavsPerkanuks, line 118: Shouldn't you check if messages already exist in DB before performing an update?
In a scenario, when db has no messages, then it will be created, which could be counterintuitive ? , because we already have createMessages endpoint

For me, it would be logically that we should not on updateRequest create resource if it does not exist. 
1

VladislavsPerkanuks: The implementation would be similar to `CreateMessages`
```go
...
if len(msgs) < 1 {
  // nothing to update
}

// get the correct messages
...
t.repo.SaveMessages(with correct messages)

```

proto/translate/v1/translate.proto
```diff
@@ -63,6 +63,15 @@ message ListMessagesResponse {
   repeated Messages messages = 1;
 }
 
+message UpdateMessagesRequest{
```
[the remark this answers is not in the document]
janishorsts: We'll have it later but not now.

pkg/translate/message.go
janishorsts: @VladislavsPerkanuks advice `len(msgs) == 0` instead of `len(msgs) < 1` as it is easier to reason with. Where is `< 1` coming from?

cmd/translate/main_grpc_integration_test.go
DarjaLeonova: I wrote rest integration tests for updateMessages and for listMessages tests were already there but with the name GetMessages

klavinsjanis (approved):

cmd/translate/main_grpc_integration_test.go
```diff
@@ -423,12 +423,12 @@ func Test_ListServices_gRPC(t *testing.T) {
 
 // ------------------Messages------------------
 
-func randMessages(t *testing.T, override *translatev1.Messages) *translatev1.Messages {
+func randMessages(t *testing.T, langOverride string) *translatev1.Messages {
```
janishorsts, line 426: why change from `*translatev1.Messages` to `string`?

VladislavsPerkanuks: in this `randMessages`, the only purpose of override was to override messages language.
so passing only language is easier than passing the whole message struct, just to get new language

DarjaLeonova: @VladislavsPerkanuks did this change and since I had a day off, I don't know the reason why it is changed

janishorsts: It's gonna go back to `override *translatev1.Messages` at one point - when `*translatev1.Messages` becomes more complex with more fields.

janishorsts (approved):

PR #110, opened 2023-06-30
Change XLIFF 1.2 and XLIFF 2.0 Import Behaviour

VladislavsPerkanuks:
## Changes

Now imported messages are determined by the attribute of the source/target language
- if target language **is not** present, then messages are extracted from `<source>` element, and `messages` language is source-language
- if target language **is** present, then messages are extracted from `<target>` element, and `messages` language is target-language

## Test changes
- Refactored tests to include less boilerplate - now `XML's` are generated dynamically from provided random messages
- Added transform tests: messages->XLIFF->messages (Property-based testing)

### DIFF
--- a/pkg/convert/xliff12.go
+++ b/pkg/convert/xliff12.go
@@ -41,7 +41,7 @@ type transUnit struct {
 func FromXliff12(data []byte) (model.Messages, error) {
 	var xlf xliff12
 	if err := xml.Unmarshal(data, &xlf); err != nil {
-		return model.Messages{}, fmt.Errorf("unmarshal to xliff12: %w", err)
+		return model.Messages{}, fmt.Errorf("unmarshal xliff12: %w", err)
 	}
 
 	messages := model.Messages{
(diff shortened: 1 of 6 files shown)


pkg/convert/xliff12.go
```diff
@@ -27,27 +30,43 @@ type bodyElement struct {
...
+		messages.Language = xlf.File.TargetLanguage
+	}
+
+	getMessage := func(t transUnit) string {
+		if isTranslated {
+			return t.Target
+		}
+
+		return t.Source
+	}
```
janishorsts, line 64: You can improve efficiency a lot. Right now, for each message it does `if` statement. The function is a first-class citizen in Golang - use it.

```suggestion
	getMessage := func(t transUnit) string { return t.Source}
	
	if isTranslated {
		getMessage = func(t transUnit) string { return t.Target }
	}
```

pkg/convert/xliff2.go
```diff
@@ -8,38 +8,60 @@ import (
...
+		Messages: make([]model.Message, 0, len(xlf.File.Units)),
+	}
+
+	// Check if file has a target language set
+	isTranslated := xlf.TrgLang != language.Und
+	if isTranslated {
+		messages.Language = xlf.TrgLang
+	}
+
+	getMessage := func(u unit) string {
```
janishorsts, line 58: As previously advised

janishorsts (changes requested):
Please merge after acting on comments!

PR #112, opened 2023-07-03
92-add-positions-to-field-message WIP

klavinsjanis:


pkg/testutil/rand/model.go
```diff
@@ -67,13 +67,19 @@ func WithID(id uuid.UUID) ModelServiceOption {
...
 		ID:          gofakeit.SentenceSimple(),
 		PluralID:    gofakeit.SentenceSimple(),
 		Message:     gofakeit.SentenceSimple(),
 		Description: gofakeit.SentenceSimple(),
 		Fuzzy:       gofakeit.Bool(),
 	}
+
+	if gofakeit.Bool() {
+		gofakeit.Slice(&msg.Positions)
+	}
```
VladislavsPerkanuks, line 80:
```suggestion
	gofakeit.Slice(&msg.Positions)
```

cmd/translate/main_grpc_integration_test.go
```diff
@@ -446,6 +446,10 @@ func randMessages(t *testing.T, override *translatev1.Messages) *translatev1.Mes
 			Fuzzy:       gofakeit.Bool(),
 		}
 
+		if gofakeit.Bool() {
+			gofakeit.Slice(message.Positions)
+		}
```
VladislavsPerkanuks, line 451:
```suggestion
		gofakeit.Slice(message.Positions)
```

migrate/mysql/000001_initial_schema.up.sql
```diff
@@ -16,6 +16,7 @@ CREATE TABLE message_message (
   id TEXT NOT NULL,
   message TEXT NOT NULL,
   description TEXT,
+  positions TEXT,
```
VladislavsPerkanuks, line 19: In smaller projects, it may be sufficient to store an extracted text in a single TEXT field if it only exists in one or two files. However, in larger projects where the same translation is likely to appear in multiple files, creating a separate table to track their positions could be a more efficient solution? 

I am just brainstorming here, maybe I am wrong :)

pkg/convert/xliff12.go
```diff
@@ -95,3 +109,59 @@ func ToXliff12(messages model.Messages) ([]byte, error) {
...
+
+// helpers
+
+// positionsFromXliff12 extracts line positions from []contextGroup.
+func positionsFromXliff12(contextGroups []contextGroup) model.Positions {
+	var positions model.Positions
+
+	for _, cg := range contextGroups {
+		if cg.Purpose == "location" {
+			var pos string
```
VladislavsPerkanuks, line 121: string builder is more efficent, as strings are immutable in Go

pkg/testutil/rand/model.go
klavinsjanis: that's there on purpose, create nil value

cmd/translate/main_grpc_integration_test.go
klavinsjanis: that's there on purpose, create nil value

migrate/mysql/000001_initial_schema.up.sql
janishorsts: The probability is zero that the value gets even close to the limit. It's good.

janishorsts: why not JSON?

klavinsjanis: for writing SQL queries to target specific source files, maybe could be useful.

klavinsjanis: I didn't really consider JSON as an option, was more on the fence about creating a new table.

pkg/model/message.go
```diff
@@ -12,5 +19,38 @@ type Message struct {
...
+		return fmt.Errorf("scan positions: %w", err)
+	}
+
+	if positions.Valid {
+		if len(positions.String) > 0 {
+			*p = strings.Split(positions.String, ", ")
+		} else {
+			*p = Positions{}
+		}
+	}
```
VladislavsPerkanuks, line 53: why empty positions and not just nil ?
```suggestion
	if positions.Valid {
		if len(positions.String) > 0 {
			*p = strings.Split(positions.String, ", ")
		}
	}
```

klavinsjanis: Tests will fail when persisting empty slice and then getting back nil from repo. The initial value will be compared to the returned one and they won't match.

pkg/convert/go.go
```diff
@@ -54,19 +56,25 @@ func messagesToPipeline(m model.Messages) pipeline.Messages {
...
+		msg := model.Message{
 			ID:          value.ID[0],
 			Fuzzy:       value.Fuzzy,
 			Description: value.Meaning,
 			Message:     convertToMessageFormatSingular(value.Message.Msg),
-		})
+		}
+
+		if value.Position != "" {
+			msg.Positions = strings.Split(value.Position, ", ")
```
janishorsts, line 73: Do `strings.Split` and a generic function for mapping - trim spaces!

pkg/convert/go_test.go
```diff
@@ -16,7 +16,11 @@ var modelMsg = model.Messages{
 			ID:          "1",
 			Message:     "{message1}",
 			Description: "description1",
-			Fuzzy:       true,
+			Positions: []string{
+				"src/config.ts:10",
```
janishorsts, line 20: golang does not use `.ts`. Use `.go`. :)

pkg/convert/xliff12.go
```diff
@@ -95,3 +109,59 @@ func ToXliff12(messages model.Messages) ([]byte, error) {
...
+						pos.WriteString(", " + c.Content)
+					} else {
+						pos.WriteString(c.Content)
+					}
+				case "linenumber":
+					pos.WriteString(":" + c.Content)
+				}
+			}
+
+			positions = append(positions, strings.Split(pos.String(), ", ")...)
```
janishorsts, line 136: Inefficient. Do not build string for all positions and then split!

Create a value for each position and then append it instead.

pkg/model/message.go
```diff
@@ -12,5 +19,38 @@ type Message struct {
 	PluralID    string
 	Message     string
 	Description string
+	Positions   Positions
 	Fuzzy       bool
 }
+
+type Positions []string
```
janishorsts, line 26: store as JSON instead of plain text

pkg/convert/go_test.go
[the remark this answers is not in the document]
klavinsjanis: Looks like it creates a new message for each.

DarjaLeonova (approved):

PR #113, opened 2023-07-03
Ensure that GO Generated Files Are Up-to-date

VladislavsPerkanuks:


.github/workflows/ci.yml
```diff
@@ -7,31 +7,32 @@ env:
   EARTHLY_VERSION: 0.7.9
 
 jobs:
-  lint:
```
VladislavsPerkanuks, line 10: I switched to a single job because parallel execution was (_likely_?) causing issues and sometimes making the integration tests fail. I hope this change will fix the problem. Additionally, the speed gains from using parallel jobs for `lint` and `test` were not significant, if any, due to the overhead of setting up Earthly and other factors.

PR #117, opened 2023-07-04
Add Original Field In Handlers, Client and Convert (2.)

VladislavsPerkanuks:
### Changes
- Added optional original flag to client
- Added check for original in xliff formats
- Updated `uploadFIleHandler` to include `Original` param
- Updated transform `messages` to include `Original` field
- Updated proto definitions with new `Original` field for `Message` and `UploadTranslationFileRequest`


cmd/client/cmd/upload.go
```diff
@@ -89,6 +95,7 @@ func newUploadCmd() *cobra.Command {
 	uploadFlags.String("language", "", "translation language")
 	uploadFlags.Var(&schemaFlag, "schema",
 		`translate schema, allowed: 'json_ng_localize', 'json_ngx_translate', 'go', 'arb', 'pot', 'xliff_12', 'xliff_2'`)
+	uploadFlags.Bool("original", false, "file's language is an original language")
```
klavinsjanis, line 98: is this an optional param?

VladislavsPerkanuks: yes

pkg/server/file.go
```diff
@@ -102,6 +103,13 @@ func (t *TranslateServiceServer) UploadTranslationFile(
 		return nil, status.Errorf(codes.InvalidArgument, err.Error())
 	}
 
+	// If after converting original is false, override it with the value from the request.
+	// For now, only XLIFF formats can determine if the file is original or not.
+	// All other format's converts marks this flag as false.
+	if messages.Original == false {
```
klavinsjanis, line 109: alternative:
`if !messages.Original`

PR #123, opened 2023-07-05
120-mark-message-as-translated

klavinsjanis:


VladislavsPerkanuks (commented):
I recommend changing the base branch to `main` to simplify the merging process and reduce branch dependencies.

There is no need to keep base branch as `118-...`

klavinsjanis:
> I recommend changing the base branch to `main` to simplify the merging process and reduce branch dependencies.
> 
> There is no need to keep base branch as `118-...` and continue branch chaining, when it possible to avoid that :)

done

pkg/testutil/rand/model.go
```diff
@@ -72,7 +72,7 @@ func modelMessage() *model.Message {
 		PluralID:    gofakeit.SentenceSimple(),
 		Message:     gofakeit.SentenceSimple(),
 		Description: gofakeit.SentenceSimple(),
-		Fuzzy:       gofakeit.Bool(),
+		Status:      model.MessageStatus(gofakeit.IntRange(0, 2)), //nolint:gomnd
```
VladislavsPerkanuks, line 75: You could create `MessageStatus() model.MessageStatus` in `testutil/rand`, I see that repeats in multiple places
then:
```suggestion
		Status:     Status()
```

proto/translate/v1/translate.proto
```diff
@@ -21,7 +21,13 @@ message Message {
...
   string message = 2;
   string description = 3;
-  bool fuzzy = 4;
+  Status status = 4;
+
+  enum Status {
+    STATUS_UNTRANSLATED = 0;
+    STATUS_FUZZY = 1;
+    STATUS_TRANSLATED = 2;
+  }
```
VladislavsPerkanuks, line 30: two things:
1. first enum element should be Unspecified
2. be consistent with other the enum (`SCHEMA`), e.g. drop the `STATUS` prefix
```suggestion
  enum Status {
    UNSPECIFIED = 0;
    UNTRANSLATED = 1;
    FUZZY = 2;
    TRANSLATED = 3;
  }
```

VladislavsPerkanuks (approved):
Overall looks goods, minor comments

DarjaLeonova (approved):

proto/translate/v1/translate.proto
klavinsjanis: do we need unspecified, isn't default - untranslated?

VladislavsPerkanuks: > do we need unspecified, isn't default - untranslated?

The UNSPECIFIED value is not required, but it is often used as a default value in case the status is not set or is unknown.

> prefixing enum values is preferred based on protobuf style guide.

we with @janishorsts  decided that was too verbose (when I defined `SCHEMA` ), and ditched the prefix. So I think it would be good to remain consistent with that in other enums as well

PR #130, opened 2023-07-18
use v3 for the google translation API

DarjaLeonova:


janishorsts:
Already implemented in another PR

PR #134, opened 2023-07-21
Update dependencies

VladislavsPerkanuks:


Earthfile
```diff
@@ -115,7 +115,7 @@ test-integration:
   COPY +go/translate /translate
   COPY --dir migrate/mysql migrate
   WITH DOCKER --compose compose.yaml --service mysql --pull migrate/migrate:v$migrate_version --pull golang:$go_version-alpine
-    RUN --no-cache \
```
VladislavsPerkanuks, line 118: If I understood correctly there is no point, in running integration tests with a disabled cache, as the token env variable is changed to `path_to_token` env variable

Drastically improves integration tests execution (only locally tho )

Earthfile
```diff
@@ -115,7 +115,7 @@ test-integration:
...
+    RUN \
       --secret=googletranslate_account_key \
```
VladislavsPerkanuks, line 119: how about mounting a secret, if we use file anyways?
```bash
RUN \
 --mount type=secret,target=/translate/google_account_key.json,id=googletranslate_account_key ...
```

the only problem, that I see is in the earthly cloud secrets token is encoded in b64

Earthfile
janishorsts: Without `--no-cache`, the google account access json is cached in the buildkit layer. We require `--no-cache` for security reasons.

janishorsts: https://github.com/translate-agent/translate/pull/134/files#diff-2705f6016a64189abd14041e2f3e8dbb829553295d9088e4eb42eccb5e61bf47R132

janishorsts (approved):

klavinsjanis (approved):

Issue #124, opened 2023-07-11
Translate using AWS Translate API


klavinsjanis:
https://docs.aws.amazon.com/translate/latest/APIReference/welcome.html

PR #136, opened 2023-07-26
Fix All In One Image And Integration Tests Failing

VladislavsPerkanuks:


VladislavsPerkanuks:
It is almost done in #124 
closing to avoid conflicts

PR #146, opened 2023-08-08
141-google-translate-codepoints-limit

klavinsjanis:


kaaposc (approved):

pkg/fuzzy/google_translate.go
```diff
@@ -16,17 +17,22 @@ import (
...
+	// error details:
+	//
+	//	name = BadRequest
+	//	field = contents
+	//	desc = Number of sub-requests should not be more than 1024
+	googleTranslateRequestLimit = 1024
+	// googleTranslateCodePointsLimit limits the number of Unicode codepoints per single translation request.
+	// https://github.com/googleapis/googleapis/blob/master/google/cloud/translate/v3/translation_service.proto#L201
+	googleTranslateCodePointsLimit = 30_000
+)
```
janishorsts, line 35:
```suggestion
// List of the google request limits based on TranslateTextRequest.contents description:
// https://github.com/googleapis/googleapis/blob/master/google/cloud/translate/v3/translation_service.proto
const (
	// googleTranslateRequestLimit limits the number of strings per translation request.
	googleTranslateRequestLimit = 1024
	
	// googleTranslateCodePointsLimit limits the number of Unicode codepoints per single translation request.
	googleTranslateCodePointsLimit = 30_000
)
```

pkg/fuzzy/google_translate.go
```diff
@@ -126,36 +132,56 @@ func (g *GoogleTranslate) Translate(
...
+			batch             []string
+			batches           [][]string
+			codePointsInBatch int
+		)
 
-		// Extract the strings to be translated by the Google Translate API.
 		for i := low; i < high; i++ {
-			req.Contents = append(req.Contents, messages.Messages[i].Message)
-		}
+			codePointsInMsg := utf8.RuneCountInString(messages.Messages[i].ID)
```
VladislavsPerkanuks, line 150: Arent we suppose to translate `message.message` not` message.id` ?

pkg/fuzzy/google_translate.go
janishorsts: Good find :), I was not able to find the limits anywhere

pkg/fuzzy/google_translate.go
janishorsts, line 150: Why `.ID`? Why not `.Message`?

klavinsjanis: bad code

pkg/fuzzy/google_translate.go
```diff
@@ -126,36 +132,56 @@ func (g *GoogleTranslate) Translate(
 		Messages: make([]model.Message, 0, n),
 	}
 
-	for low := 0; low < n; low += googleTranslateRequestlimit {
-		high := low + googleTranslateRequestlimit
+	for low := 0; low < n; low += googleTranslateRequestLimit {
```
janishorsts, line 135: Can you simplify it? Can you use the two constraints (limits) as a threshold for batching?

pkg/fuzzy/google_translate.go
```diff
@@ -119,43 +118,60 @@ func (g *GoogleTranslate) Translate(
...
 	}
 
-	for low := 0; low < n; low += googleTranslateRequestlimit {
-		high := low + googleTranslateRequestlimit
+	// Split text from messages into batches to avoid exceeding
+	// googleTranslateRequestLimit or googleTranslateCodePointsLimit.
+
+	var codePointsInBatch, msgsInBatch int
+
+	batch := make([]string, 0)
```
janishorsts, line 132: What is the benefit of a zero-length empty slice over a nil slice? Can we predict the required size?

klavinsjanis: Lint was complaining about pre-alloc. We could pre-allocate capacity using translateRequestLimit, but mostly I think batch size will be less, about 600.

janishorsts: We know, for sure, the minimum is at least one, and also the maximum size is known - googleTranslateRequestLimit. We can reserve the upper limit, we may not use all reserved memory but the size is quite small.

```suggestion
	batch := make([]string, 0, googleTranslateRequestLimit)
```

pkg/fuzzy/google_translate.go
```diff
@@ -119,43 +118,59 @@ func (g *GoogleTranslate) Translate(
 		return &model.Messages{Language: messages.Language, Original: messages.Original}, nil
 	}
 
-	n := len(messages.Messages)
 	translatedMessages := model.Messages{
```
janishorsts, line 121: Why translatedMessages is so far away from its consumption now?

pkg/fuzzy/google_translate.go
```diff
@@ -119,43 +118,59 @@ func (g *GoogleTranslate) Translate(
...
-	n := len(messages.Messages)
+	// Split text from messages into batches to avoid exceeding
+	// googleTranslateRequestLimit or googleTranslateCodePointsLimit.
+
+	var codePointsInBatch int
+
+	batch := make([]string, 0, googleTranslateRequestLimit)
+	batches := make([][]string, 0, 1)
+
+	for _, msgs := range messages.Messages {
```
janishorsts, line 129: Why do you use `msgs` name? Why not `msg` or `message` or `v`?

pkg/fuzzy/google_translate.go
```diff
@@ -119,43 +118,59 @@ func (g *GoogleTranslate) Translate(
...
-		for i := low; i < high; i++ {
-			req.Contents = append(req.Contents, messages.Messages[i].Message)
-		}
-
-		res, err := g.client.TranslateText(ctx, req)
+			Contents:           batches[i],
+		})
 		if err != nil {
-			return nil, fmt.Errorf("google translate client: translate text: %w", err)
+			return nil, fmt.Errorf("google translate client: translate text from batch at index %d: %w", i, err)
```
janishorsts, line 163:
```suggestion
			return nil, fmt.Errorf("google translate client: translate texts from batch #%d: %w", i, err)
```

janishorsts (approved):

PR #147, opened 2023-08-10
142-badgerdb-exceeding-size-of-value-limit

klavinsjanis:


cmd/translate/service/service.go
```diff
@@ -71,9 +71,29 @@ var rootCmd = &cobra.Command{
 		// Gracefully stops GRPC server.
 		defer grpcServer.GracefulStop()
 
+		dbType := viper.GetString("service.db")
+		badgerDir := viper.GetString("db.badgerdb.path")
+
+		if dbType == repo.BadgerDB && badgerDir == "" {
```
VladislavsPerkanuks, line 77: I think more clean solution would be just making flag/env `db.badgerdb.path` mandatory

```go
func WithDefaultDB() option {
  ...
  if path == "" {
    return errors.New("No path set")
  }
  ...
```

klavinsjanis: Well, the downside of that will be that the path will be mandatory.

VladislavsPerkanuks: ¯\\_(ツ)_/¯

VladislavsPerkanuks: I just want to facilitate this function as much as possible

kaaposc (approved):

DarjaLeonova (approved):

cmd/translate/service/service.go
janishorsts: The context of how the service is run guides us. It is always packaged as an image.

In an image, we have a predefined path where we store mutable data. Thereof, we can set a default path - e.g. `/data/badgerdb`. When the service is run in a container, the mutable data can be mounted for persistent storage.

I understand TMPDIR might be helpful for local development.

What does viper do if the value is empty and it has a default value defined?

pkg/repo/badgerdb/repo.go
```diff
@@ -23,6 +23,9 @@ func WithDefaultDB() option {
 		path := viper.GetString("db.badgerdb.path")
 		badgerOpts := badger.DefaultOptions(path)
 
+		// NOTE: The default value for in-memory storage of ValueThreshold is 1 MB.
```
janishorsts, line 26: Indeed, it is constrained by badger.maxValueThreshold https://github.com/dgraph-io/badger/blob/v3.2103.5/options.go#L210

cmd/translate/service/service.go
klavinsjanis: It will use the default value that was set. Maybe we could set the badgerdb path as e.g. `/data/badgerdb` when we build the all-in-one image.

janishorsts: In principle, the defaults must be for the "production" environment which is a container. We need to ensure all defaults follow it and defaults are consistent.

I like the idea of a temporary folder for local development or a CD/CI environment. The challenge is how we configure it. Do we set `db.badgerdb.path` to `TMPDIR` to create a temporary directory when working on the project locally?

cmd/translate/service/service.go
```diff
@@ -71,9 +71,25 @@ var rootCmd = &cobra.Command{
 		// Gracefully stops GRPC server.
 		defer grpcServer.GracefulStop()
 
+		dbType := viper.GetString("service.db")
+		badgerDir := viper.GetString("db.badgerdb.path")
+		badgerTmpDir := os.TempDir() + "badgerdb"
```
janishorsts, line 76: These are executed even when another db is configured to be used.

cmd/translate/service/service.go
```diff
@@ -71,9 +71,25 @@ var rootCmd = &cobra.Command{
...
+			log.Printf("INFO: Using temporary storage directory '%s' for badgerdb\n", badgerTmpDir)
+
+			defer func() {
+				if rmErr := os.RemoveAll(badgerDir); rmErr != nil {
+					log.Printf("WARNING: Failed to remove temporary storage directory '%s' for badgerdb: %v\n", badgerTmpDir, rmErr)
+					return
+				}
+				log.Printf("INFO: Deleting temporary storage directory '%s' for badgerdb\n", badgerTmpDir)
+			}()
+		}
```
janishorsts: We don't require to remove the temporary directory at the end. All tests MUST be isolated anyway.

PR #149, opened 2023-08-10
added original field to all converting From functions

DarjaLeonova:


pkg/convert/xliff12.go
```diff
@@ -50,7 +50,7 @@ type context struct {
 }
 
 // FromXliff12 converts serialized data from the XML data in the XLIFF 1.2 format into a model.Messages struct.
-func FromXliff12(data []byte) (model.Messages, error) {
+func FromXliff12(data []byte, original bool) (model.Messages, error) {
```
DarjaLeonova, line 53: I'm not sure how to use original param in xlliff formats because it is already in use, so should we ignore it?

pkg/convert/arb_test.go
```diff
@@ -175,14 +186,15 @@ func Test_FromArb(t *testing.T) {
...
+			original:    false,
 		},
 	}
 	for _, tt := range tests {
 		tt := tt
 		t.Run(tt.name, func(t *testing.T) {
 			t.Parallel()
 
-			actual, err := FromArb(tt.input)
+			actual, err := FromArb(tt.input, tt.original)
```
janishorsts, line 197: Keep it simple! No need to declare `original` variable.

```suggestion
			actual, err := FromArb(tt.input, false)
```

pkg/convert/json_nglocalize_test.go
```diff
@@ -54,14 +57,15 @@ func Test_FromNgLocalize(t *testing.T) {
...
+			original:    false,
 		},
 	}
 	for _, tt := range tests {
 		tt := tt
 		t.Run(tt.name, func(t *testing.T) {
 			t.Parallel()
 
-			actual, err := FromNgLocalize(tt.input)
+			actual, err := FromNgLocalize(tt.input, tt.original)
```
janishorsts, line 68: the same here, no need for `original`, simply literal value `false`.

pkg/convert/arb_test.go
DarjaLeonova: Oki, I will change then in another tests too  ;)

kaaposc (approved):

pkg/convert/xliff12.go
VladislavsPerkanuks: I think there is sufficient info to determine the originality of the file, e.g. there is no target language in file. So original parameter could be ignored. 
```suggestion
func FromXliff12(data []byte, _ bool) (model.Messages, error) {
```

VladislavsPerkanuks (changes requested):
Remove this, as it is not needed anymore
https://github.com/translate-agent/translate/blob/bd37c070144b2749c062480424a6824703cbd3df/pkg/server/file.go#L112C1-L117C3

pkg/convert/xliff12.go
janishorsts: Let's keep `original` name for clarity and intent. It might be required to process xliff in the future.

janishorsts: @DarjaLeonova can you please add a comment to all `From...` where `original` is ignored?

DarjaLeonova: sure, I will add now

pkg/convert/go.go
```diff
@@ -23,14 +23,15 @@ func ToGo(m model.Messages) ([]byte, error) {
 
 // FromGo takes a JSON-encoded byte slice, decodes it into a pipeline.Messages structure,
 // and then converts it into a model.Messages structure using the messagesFromPipeline function.
-func FromGo(b []byte) (model.Messages, error) {
+// For now original param is ignored.
```
klavinsjanis, line 26: is it ignored or are you using it in the function?

klavinsjanis (approved):

janishorsts (approved):

kaaposc (approved):

PR #152, opened 2023-08-10
151-correct-endpoint-for-updating-messages

klavinsjanis:


proto/translate/v1/translate.proto
```diff
@@ -151,7 +152,7 @@ service TranslateService {
 
   rpc UpdateMessages(UpdateMessagesRequest) returns (Messages) {
     option (google.api.http) = {
-      put: "/v1/services/{service_id}/messages"
+      put: "/v1/services/{service_id}/messages/{language}"
```
janishorsts, line 155: Does proto allow the following?

```suggestion
      put: "/v1/services/{service_id}/messages/{messages.language}"
```

klavinsjanis: It is possible but might cause issues when making requests using REST client. By providing different language in the path  messages.language will be overwritten.

janishorsts: I see. The issue is an incorrectly constructed REST request (bad request) processed successfully when the duplicate field is not equal. On the opposite side, it degrades the gRPC request by adding the duplicate field. In our case, gRPC is the primary interface, we should keep it clean. Let's use `{messages.language}`! REST API is nice to have anyway.

kaaposc (approved):

janishorsts (approved):

PR #153, opened 2023-08-11
133-mark-translations-as-fuzzy-when-original-text-message-changes-but-key-stays-the-same

klavinsjanis:


pkg/server/message.go
```diff
@@ -236,18 +233,117 @@ func (t *TranslateServiceServer) UpdateMessages(
...
+	}
+
+	var originalMessages *model.Messages
+
+	for i := range allMessages {
+		if allMessages[i].Original {
+			originalMessages = &allMessages[i]
+			break
+		}
+	}
```
VladislavsPerkanuks, line 277: I am thinking about making `[]model.Messages `a type, and this could be one of its methods. 
On the upside cleaner code, as the downside, much refactoring is needed, almost everywhere...

kaaposc (approved):
Alternative: please consider returning a new messages array instead of changing them in-place in function `updateAlteredMessages`.

pkg/server/file.go
```diff
@@ -130,6 +135,22 @@ func (t *TranslateServiceServer) UploadTranslationFile(
...
+		// if populateTranslations flag is true, populate the translated messages.
+		if params.populateTranslations {
+			if err = t.populateTranslatedMessages(ctx, params.serviceID, messages, allMessages); err != nil {
+				return nil, err
+			}
+		}
+	}
+
 	// If the uploaded file is original and populateTranslations flag is true, populate the translated messages.
 	if messages.Original && params.populateTranslations {
```
VladislavsPerkanuks, line 155: Isn't this redundant, now ?

pkg/server/file.go
```diff
@@ -130,6 +135,22 @@ func (t *TranslateServiceServer) UploadTranslationFile(
 		return nil, status.Errorf(codes.Internal, "")
 	}
 
+	if messages.Original && len(allMessages) != 0 {
+		// TODO: optimize performance when populateTranslations param is true, currently saveMessages() will be called twice.
```
VladislavsPerkanuks, line 139: you could change to` t.updateAlteredMessages` and ` t.populateTranslatedMessages` to return an `[]model.Messages`, and then in handler call `SaveMessages`, in one for loop

As another improvement, we could define a new method for repos, `SaveMessagesSlice`, or something like that, which takes an `[]model.Messages` as argument

pkg/server/message.go
```diff
@@ -236,18 +233,117 @@ func (t *TranslateServiceServer) UpdateMessages(
...
+	originalMessagesLookup := make(map[string]*model.Message, len(originalMessages.Messages))
+
+	for i := range originalMessages.Messages {
+		originalMessagesLookup[originalMessages.Messages[i].ID] = &originalMessages.Messages[i]
+	}
+
+	// find altered original messages
+	alteredMessages := model.Messages{
+		Language: newOriginalMessages.Language,
+		Original: true,
```
VladislavsPerkanuks, line 294: this is on the question mark, but how about reserving memory for the length of original messages `len(originalMessages.Messages)`, instead of letting it grow automatically?

klavinsjanis: What if there are thousands of messages, but only one of them is altered?

pkg/server/file.go
klavinsjanis: yeah there are a lot of possible improvements, wanted avoid any scope creep.

pkg/server/file.go
```diff
@@ -121,6 +121,11 @@ func (t *TranslateServiceServer) UploadTranslationFile(
 		return nil, status.Errorf(codes.InvalidArgument, err.Error())
 	}
 
+	allMessages, err := t.repo.LoadMessages(ctx, params.serviceID, common.LoadMessagesOpts{})
```
janishorsts, line 124: The flow of uploading translation file can be simplified a lot. Why do we have two loads, two saves, two populated messages etc?

klavinsjanis: forgot to remove the previous populate messages

pkg/server/message.go
VladislavsPerkanuks: Depends on the frontend, if it calls `updateMessages` on every translated messages, then yes, but if it is called only once, after the translator, is finished translating some batch, then that would be another story

pkg/model/message.go
```diff
@@ -14,6 +14,29 @@ type Messages struct {
...
+			Language: m[i].Language,
+			Messages: make([]Message, len(m[i].Messages)),
+			Original: m[i].Original,
+		}
+
+		copy(msgs[i].Messages, m[i].Messages)
+	}
+
+	return msgs
+}
```
VladislavsPerkanuks, line 38: As we don't have any references inside the Messages struct, we can make a shallow copy.

And shallow copy can be made using standard lib (go 1.21) slices.Clone

So this method becomes redundant

klavinsjanis: We have a reference to the messages slice, it will get modified if shallow copy is used.

pkg/server/message.go
```diff
@@ -236,47 +233,156 @@ func (t *TranslateServiceServer) UpdateMessages(
...
+	}
+
+	for _, newMessage := range newOriginalMessages.Messages {
+		if message, ok := originalMessagesLookup[newMessage.ID]; ok {
+			if message.Message != newMessage.Message {
```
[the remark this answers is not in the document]
klavinsjanis: hmm, then I would have to check if the message is nil before I could compare the text.

pkg/model/message.go
```diff
@@ -14,6 +14,44 @@ type Messages struct {
 	Original bool
 }
 
+type MessagesSlice []Messages
+
+// GetCopy() returns a deep copy of MessagesSlice.
+func (m MessagesSlice) GetCopy() MessagesSlice {
+	if m == nil {
```
janishorsts, line 21: This is not a correct expression for a slice for your use case. Which one is?

janishorsts: Also, the receiver is passed by a value, not a pointer, `m == nil` is never true.

pkg/model/message.go
```diff
@@ -14,6 +14,44 @@ type Messages struct {
 	Original bool
 }
 
+type MessagesSlice []Messages
+
+// GetCopy() returns a deep copy of MessagesSlice.
+func (m MessagesSlice) GetCopy() MessagesSlice {
```
janishorsts, line 20:
```suggestion
func (m MessagesSlice) Clone() MessagesSlice {
```

pkg/model/message.go
```diff
@@ -14,6 +14,44 @@ type Messages struct {
...
+
+		copy(msgs[i].Messages, m[i].Messages)
+	}
+
+	return msgs
+}
+
+// GetOriginal returns a pointer to original messages.
+func (m MessagesSlice) GetOriginal() *Messages {
+	if m == nil {
```
janishorsts, line 42: incorrect experssion

pkg/model/message.go
klavinsjanis: len() -> if empty slice is provided, new empty slice is returned, but I guess this check can be skipped.

pkg/model/message.go
```diff
@@ -14,6 +14,36 @@ type Messages struct {
 	Original bool
 }
 
+type MessagesSlice []Messages
+
+// Clone() returns a deep copy of MessagesSlice.
+func (m MessagesSlice) Clone() MessagesSlice {
+	msgs := make(MessagesSlice, len(m))
```
janishorsts, line 21: Let's return a nil slice instead of an empty slice when `len(m) == 0`.

```suggestion
	msgs := make(MessagesSlice, len(m))
```
```suggestion
	if len(m) == 0 {
		nil
	}

	msgs := make(MessagesSlice, len(m))
```

pkg/model/message.go
```diff
@@ -14,6 +14,36 @@ type Messages struct {
...
+
+	for i := range m {
+		msgs[i] = Messages{
+			Language: m[i].Language,
+			Messages: make([]Message, len(m[i].Messages)),
+			Original: m[i].Original,
+		}
+
+		copy(msgs[i].Messages, m[i].Messages)
+	}
```
janishorsts, line 31:
```suggestion
	copy(msgs, m)

	for i := range m {
		copy(msgs[i].Messages, m[i].Messages)
	}
```

pkg/server/message.go
```diff
@@ -214,15 +214,12 @@ func (t *TranslateServiceServer) UpdateMessages(
 		return nil, status.Errorf(codes.InvalidArgument, err.Error())
 	}
 
-	msgs, err := t.repo.LoadMessages(
-		ctx,
-		params.serviceID,
-		common.LoadMessagesOpts{FilterLanguages: []language.Tag{}})
+	allMessages, err := t.repo.LoadMessages(ctx, params.serviceID, common.LoadMessagesOpts{FilterLanguages: []language.Tag{}})
```
janishorsts, line 217:
```suggestion
	allMessages, err := t.repo.LoadMessages(ctx, params.serviceID, common.LoadMessagesOpts{})
```

pkg/model/message.go
```diff
@@ -14,6 +14,36 @@ type Messages struct {
...
+// GetOriginal returns a pointer to original messages.
+func (m MessagesSlice) GetOriginal() *Messages {
+	for i := range m {
+		if m[i].Original {
+			return &m[i]
+		}
+	}
+
+	return nil
+}
```
janishorsts, line 45: Finding original and other messages might be more common behaviour. This also improves the consumption in your code base.

```suggestion
// SplitOriginal returns a pointer to the original and other messages.
func (m MessagesSlice) SplitOriginal() (original *Messages, others MessagesSlice) {
}
```

pkg/server/message.go
janishorsts: Why `Original: true,`?

```suggestion
```

klavinsjanis: umm, just added it for additional clarification that those are original messages.

pkg/server/message.go
```diff
@@ -236,47 +233,147 @@ func (t *TranslateServiceServer) UpdateMessages(
...
+		if messages.Original {
+			continue
+		}
+
+		translated, err := t.translator.Translate(ctx, &alteredMessages, messages.Language)
+		if err != nil {
+			return nil, status.Errorf(codes.Unknown, err.Error()) // TODO: For now we don't know the cause of the error.
+		}
+
+		translatedMessagesLookup := make(map[string]*model.Message, len(translated.Messages))
```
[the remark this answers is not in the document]
klavinsjanis: That's a pointer to the message struct.

pkg/model/message.go
```diff
@@ -14,6 +14,45 @@ type Messages struct {
...
+	others = m
+
+	for i := range m {
+		if m[i].Original {
+			original = &m[i]
+
+			if len(m) > 2 { //nolint:gomnd
+				others[i] = others[len(m)-1]
+				others = others[:len(m)-1]
+			}
```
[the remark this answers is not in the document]
klavinsjanis: ok, this is not correct

pkg/server/file.go
janishorsts, line 1: The flow is inefficient. Both `alterTranslations()` and `populateTranslations()` do fuzzy translations. We should do fuzzy translation only once.

pkg/server/file.go
```diff
@@ -123,16 +123,47 @@ func (t *TranslateServiceServer) UploadTranslationFile(
...
 		allMessages, err := t.repo.LoadMessages(ctx, params.serviceID, repo.LoadMessagesOpts{})
 		if err != nil {
 			return nil, status.Errorf(codes.Internal, "")
 		}
 
-		if err = t.populateTranslatedMessages(ctx, params.serviceID, messages, allMessages); err != nil {
+		// Find original messages with altered text, then replace text in associated messages for all translations.
+		newMessages, err := t.alterTranslations(ctx, allMessages, messages)
+		if err != nil {
+			return nil, err
```
VladislavsPerkanuks, line 136: error wrapping ?
In other places too

pkg/server/message.go
```diff
@@ -242,76 +239,144 @@ func (t *TranslateServiceServer) UpdateMessages(
...
 
-	// If updating the original messages and populateTranslations flag is true, update all translated messages as well.
-	if params.messages.Original && params.populateTranslations {
-		if err := t.populateTranslatedMessages(ctx, params.serviceID, params.messages, msgs); err != nil {
+	// When updating original messages, changes might affect translations - transform and update all translations.
+	if params.messages.Original {
+		// Find original messages with altered text, then replace text in associated messages for all translations.
+		newMessages, err := t.alterTranslations(ctx, allMessages, params.messages)
+		if err != nil {
 			return nil, err
```
VladislavsPerkanuks, line 247: error wrapping ?
In other places too

pkg/server/file.go
```diff
@@ -123,15 +123,46 @@ func (t *TranslateServiceServer) UploadTranslationFile(
...
+
+		// If populateMessages is true - populate missing messages for all translations.
+		if params.populateTranslations {
+			if newMessages, err = t.populateTranslations(ctx, newMessages, messages); err != nil {
+				return nil, fmt.Errorf("populate missing messages :%w", err)
+			}
+		}
+
+		// Fuzzy translate untranslated messages for all translations
+		if newMessages, err = t.refreshTranslations(ctx, newMessages, messages); err != nil {
```
janishorsts, line 147:
```suggestion
		if newMessages, err = t.fuzzyTranslate(ctx, newMessages, messages); err != nil {
```

pkg/server/file.go
```diff
@@ -123,15 +123,46 @@ func (t *TranslateServiceServer) UploadTranslationFile(
...
+	if messages.Original {
 		allMessages, err := t.repo.LoadMessages(ctx, params.serviceID, repo.LoadMessagesOpts{})
 		if err != nil {
 			return nil, status.Errorf(codes.Internal, "")
 		}
 
-		if err = t.populateTranslatedMessages(ctx, params.serviceID, messages, allMessages); err != nil {
-			return nil, err
+		// Find original messages with altered text, then replace text in associated messages for all translations.
+		newMessages, err := t.alterTranslations(ctx, allMessages, messages)
```
janishorsts, line 134:
```suggestion
		allMessages, err = t.alterTranslations(ctx, allMessages, messages)
```

pkg/server/file.go
```diff
@@ -123,15 +123,46 @@ func (t *TranslateServiceServer) UploadTranslationFile(
...
+		if err != nil {
+			return nil, fmt.Errorf("replace text in associated messages :%w", err)
+		}
+
+		// If populateMessages is true - populate missing messages for all translations.
+		if params.populateTranslations {
+			if newMessages, err = t.populateTranslations(ctx, newMessages, messages); err != nil {
```
janishorsts, line 141:
```suggestion
			if allMessages, err = t.populateTranslations(ctx, allMessages, messages); err != nil {
```

pkg/server/file.go
```diff
@@ -123,15 +123,46 @@ func (t *TranslateServiceServer) UploadTranslationFile(
...
+				return nil, fmt.Errorf("populate missing messages :%w", err)
+			}
+		}
+
+		// Fuzzy translate untranslated messages for all translations
+		if newMessages, err = t.refreshTranslations(ctx, newMessages, messages); err != nil {
+			return nil, fmt.Errorf("translate untranslated messages :%w", err)
```
janishorsts, line 148:
```suggestion
			return nil, fmt.Errorf("fuzzy translate messages :%w", err)
```

pkg/server/message.go
```diff
@@ -254,64 +282,101 @@ func (t *TranslateServiceServer) UpdateMessages(
...
 
-// populateTranslatedMessages adds any missing messages from the original messages to all translated messages.
-func (t *TranslateServiceServer) populateTranslatedMessages(
+// alterTranslations replaces text in messages for all translations
+// based on changes in the original messages text, returns messagesSlice containing altered translations.
+func (t *TranslateServiceServer) alterTranslations(
+	ctx context.Context,
+	allMessages model.MessagesSlice,
+	newOriginalMessages *model.Messages,
+) (model.MessagesSlice, error) {
```
janishorsts, line 291: Let the caller replace original Messages in allMessages and pass it to `alterTranslations()`.

```suggestion
func (t *TranslateServiceServer) alterTranslations(
	ctx context.Context,
	allMessages model.MessagesSlice,
) (model.MessagesSlice, error) {
```

pkg/server/message.go
```diff
@@ -254,64 +282,101 @@ func (t *TranslateServiceServer) UpdateMessages(
...
+// returns messagesSlice containing populated translations.
+func (t *TranslateServiceServer) populateTranslations(
 	ctx context.Context,
-	serviceID uuid.UUID,
-	originalMessages *model.Messages,
-	allMessages []model.Messages,
-) error {
+	allMessages model.MessagesSlice,
+	newOriginalMessages *model.Messages,
+) ([]model.Messages, error) {
```
janishorsts, line 344: The same here, caller replaces original Messages in allMessages and passes it to this function.

pkg/server/message.go
```diff
@@ -320,3 +385,56 @@ func langExists(msgs []model.Messages, lang language.Tag) bool {
...
 	})
 }
+
+// refreshTranslations fuzzy translates any untranslated messages,
+// returns messagesSlice containing refreshed translations.
+func (t *TranslateServiceServer) refreshTranslations(
+	ctx context.Context,
+	allMessages model.MessagesSlice,
+	newOriginalMessages *model.Messages,
+) ([]model.Messages, error) {
```
janishorsts, line 395:
```suggestion
func (t *TranslateServiceServer) refreshTranslations(
	ctx context.Context,
	allMessages model.MessagesSlice,
) ([]model.Messages, error) {
```

pkg/server/file.go
```diff
@@ -123,15 +123,44 @@ func (t *TranslateServiceServer) UploadTranslationFile(
...
+
+			err = t.repo.SaveMessages(ctx, params.serviceID, &all[i])
+			switch {
+			default:
+				// noop
+			case errors.Is(err, repo.ErrNotFound):
+				return nil, status.Errorf(codes.NotFound, "service not found")
+			case err != nil:
+				return nil, status.Errorf(codes.Internal, "")
+			}
```
VladislavsPerkanuks, line 163: I am almost sure that it is possible to move this from `if messages.Original` block, with minor changes on outside block, thus reducing repetition

pkg/server/message.go
```diff
@@ -242,10 +239,39 @@ func (t *TranslateServiceServer) UpdateMessages(
...
-	if params.messages.Original && params.populateTranslations {
-		if err := t.populateTranslatedMessages(ctx, params.serviceID, params.messages, msgs); err != nil {
-			return nil, err
+	// When updating original messages, changes might affect translations - transform and update all translations.
+	if params.messages.Original {
+		all = all.Replace(*params.messages)
+		prev, _ := all.SplitOriginal()
+
+		// Find original messages with altered text, then replace text in associated messages for all translations.
+		all = t.alterTranslations(all, findUntranslatedIDs(*prev, *params.messages))
```
VladislavsPerkanuks, line 248: same thing about dereferencing

pkg/server/message.go
```diff
@@ -254,69 +280,154 @@ func (t *TranslateServiceServer) UpdateMessages(
...
+
+	// Return if messages don't contain original language
+	if original == nil {
+		return all
+	}
+
+	// Update altered messages for all translations
+	for _, msg := range others {
+		for i := range msg.Messages {
+			if slices.Contains(untranslatedIDs, msg.Messages[i].ID) {
```
janishorsts, line 310: outside `for` we can sort `untranslatedIDs`, afterwards we can use binary search to determine if the slice contains the value

```go
slices.Sort(untranslatedIDs)
```

```suggestion
			if _, found := slices.BinarySearch(untranslatedIDs, msg.Messages[i].ID); found {
```

pkg/server/message.go
```diff
@@ -254,69 +280,154 @@ func (t *TranslateServiceServer) UpdateMessages(
...
-	originalMessages *model.Messages,
-	allMessages []model.Messages,
-) error {
+// alterTranslations alter the status of specific messages in multiple languages to "untranslated"
+// based on a list of message IDs, while keeping the "original" language messages unchanged.
+func (t *TranslateServiceServer) alterTranslations(
+	all model.MessagesSlice,
+	untranslatedIDs []string,
+) model.MessagesSlice {
+	all = all.Clone()
```
janishorsts, line 289: Let's delete `Clone()` - it is a fragile function.

pkg/server/message.go
```diff
@@ -254,69 +280,154 @@ func (t *TranslateServiceServer) UpdateMessages(
...
+
+	if len(untranslatedIDs) == 0 {
+		return all
+	}
+
+	// Return if only one language is present
+	if len(all) == 1 {
+		return all
+	}
```
janishorsts, line 298: alternatively

```suggestion
	if len(untranslatedIDs) == 0 || len(all) == 1 {
		return all
	}
```

pkg/server/message.go
```diff
@@ -242,10 +239,39 @@ func (t *TranslateServiceServer) UpdateMessages(
...
+		}
+
+		// Fuzzy translate untranslated messages for all translations
+		if all, err = t.fuzzyTranslate(ctx, all); err != nil {
+			return nil, fmt.Errorf("fuzzy translate messages: %w", err)
+		}
+
+		// Update messages for all translations
+		for i := range all {
+			if all[i].Original {
```
janishorsts, line 262: How can we avoid checking `Original` in `for`?

pkg/server/message.go
```diff
@@ -254,69 +280,154 @@ func (t *TranslateServiceServer) UpdateMessages(
...
+		}
+	}
+
+	return all
+}
+
+// populateTranslations adds any missing messages from the original messages to all translated messages,
+// returns messagesSlice containing populated translations.
+func (t *TranslateServiceServer) populateTranslations(all model.MessagesSlice) []model.Messages {
+	newMsgs := all.Clone()
```
janishorsts, line 322: delete

```suggestion
```

pkg/server/file.go
```diff
@@ -152,21 +151,28 @@ func (t *TranslateServiceServer) UploadTranslationFile(
...
+	default:
+		// noop
+	case errors.Is(err, repo.ErrNotFound):
+		return nil, status.Errorf(codes.NotFound, "service not found"), true
+	case err != nil:
+		return nil, status.Errorf(codes.Internal, ""), true
+	}
+	return nil, nil, false
+}
+
```
VladislavsPerkanuks, line 175: I think this function is not needed, as it does not benefit code readability, at least in its current state.
There are many things wrong with that:
1. **Function Naming**: Functions must do one thing, e.g. **Single Responsibility Principle**, and when you use `and` in the function name, it usually implies that it breaks that principle
2. For return values, only `error` would be enough -> `func(...) error {}`
3. No reason to ignore `default` case, and then executing `default` statement outside `switch` statement

pkg/server/file.go
```diff
@@ -114,30 +114,65 @@ func (t *TranslateServiceServer) UploadTranslationFile(
...
+			return nil, fmt.Errorf("fuzzy translate messages :%w", err)
+		}
+
+		// Update all translations
+		for i := range all {
+			if all[i].Original {
+				continue
+			}
+
+			empty, err, done = t.SaveMsgsAndHandleErrors(ctx, params, messages)
```
janishorsts, line 154: There is no need for such a simple function. Any reason you added it? Cognitive complexity?

pkg/server/message.go
```diff
@@ -258,26 +254,30 @@ func (t *TranslateServiceServer) UpdateMessages(
...
+	switch {
+	default:
+		// noop
+	case errors.Is(err, repo.ErrNotFound):
+		return nil, status.Errorf(codes.NotFound, "service not found"), true
+	case err != nil:
+		return nil, status.Errorf(codes.Internal, ""), true
+	}
+	return nil, nil, false
+}
```
VladislavsPerkanuks, line 279: almost the same as `SaveMsgsAndHandleErrors` ?
There must be a way to combine them

pkg/server/file.go
janishorsts: I can see it is reused in a couple of places.

pkg/server/message.go
```diff
@@ -307,7 +300,7 @@ func (t *TranslateServiceServer) alterTranslations(
 	// Update altered messages for all translations
 	for _, msg := range others {
 		for i := range msg.Messages {
-			if slices.Contains(untranslatedIDs, msg.Messages[i].ID) {
+			if _, found := slices.BinarySearch(untranslatedIDs, msg.Messages[i].ID); found {
```
VladislavsPerkanuks, line 303: BinarySearch works only on sorted arrays/slices. I do not see any sorting🤔

pkg/server/message.go
```diff
@@ -220,103 +220,203 @@ func (t *TranslateServiceServer) UpdateMessages(
...
-		ctx,
-		params.serviceID,
-		repo.LoadMessagesOpts{FilterLanguages: []language.Tag{}})
+	all, err := t.repo.LoadMessages(ctx, params.serviceID, repo.LoadMessagesOpts{})
 	if err != nil {
 		return nil, status.Errorf(codes.Internal, "")
 	}
 
-	if len(msgs) == 0 || !langExists(msgs, params.messages.Language) {
+	if len(all) == 0 || !langExists(all, params.messages.Language) {
```
janishorsts, line 228:
```suggestion
	if !langExists(all, params.messages.Language) {
```

VladislavsPerkanuks:
Fix for`earthly +up` not working on the first try.

in `compose.yaml`
```yaml
services:
  mysql:
    ...
    healthcheck:
      test: ["CMD", "mysqladmin" ,"ping", "-h", "localhost" ,"--wait=30"]
```

pkg/server/message.go
```diff
@@ -254,69 +271,145 @@ func (t *TranslateServiceServer) UpdateMessages(
...
+			}
+		}
+	}
+
+	return all
+}
+
+// populateTranslations adds any missing messages from the original messages to all translated messages,
+// returns messagesSlice containing populated translations.
+func (t *TranslateServiceServer) populateTranslations(all model.MessagesSlice) []model.Messages {
```
janishorsts, line 307: Does this function work? The logic in this function is not correct, the lookup contains the same messages as in the `for` loop. It compares to the same `Messages` structure.

pkg/server/file.go
```diff
@@ -114,22 +114,24 @@ func (t *TranslateServiceServer) UploadTranslationFile(
...
 
-	empty, err, done := t.SaveMsgsAndHandleErrors(ctx, params, messages)
-	if done {
-		return empty, err
+	_, err = t.repo.LoadService(ctx, params.serviceID)
+	switch {
+	case errors.Is(err, repo.ErrNotFound):
+		return nil, status.Errorf(codes.NotFound, "service not found")
+	case err != nil:
+		return nil, status.Errorf(codes.Internal, "")
```
VladislavsPerkanuks, line 122: expensive as redundant?, as `repo.SaveMessages` returns error if no `serviceID` is found, and `repo.LoadMessages` returns empty slice

DarjaLeonova: Yes, correct. Will fix it, thanks ;)

janishorsts: My bad. We need to focus on the happy path more as it delivers much higher value than chasing insignificant issues like "not found" - Pareto principle.

pkg/server/message.go
```diff
@@ -305,39 +305,27 @@ func (t *TranslateServiceServer) alterTranslations(
...
+				return message.ID == original.Messages[i].ID
+			})
+			if !found {
+				others[j].Messages = append(others[j].Messages, model.Message{
+					ID: original.Messages[i].ID,
+					Message: original.Messages[i].Message,
+					PluralID: original.Messages[i].PluralID,
+					Description: original.Messages[i].Description,
+					Positions: original.Messages[i].Positions,
+					Status: model.MessageStatusUntranslated,
```
VladislavsPerkanuks, line 322: I think this also should work 
```suggestion
				others[j].Messages = append(others[j].Messages, original.Messages[i])
```

VladislavsPerkanuks: but, there should be a status change to `untranslated`...
So idk, if it is better then

DarjaLeonova: I think for now we can keep it as it is

pkg/server/message.go
VladislavsPerkanuks: Tests were way off :)
Fixed the tests, and now I am sure that everything is working

pkg/model/message.go
```diff
@@ -14,6 +15,64 @@ type Messages struct {
...
+
+			return m
+		}
+	}
+
+	return append(m, messages)
+}
+
+// Clone returns a deep copy of MessagesSlice.
+func (m MessagesSlice) Clone() MessagesSlice {
```
janishorsts, line 47: delete?

pkg/model/message.go
```diff
@@ -14,6 +15,64 @@ type Messages struct {
 	Original bool
 }
 
+type MessagesSlice []Messages
+
+// HasLanguage checks if MessagesSlice contains Messages with the given language.
+func (ms MessagesSlice) HasLanguage(targetLang language.Tag) bool {
```
janishorsts, line 21:
```suggestion
func (ms MessagesSlice) HasLanguage(lang language.Tag) bool {
```

pkg/model/message.go
```diff
@@ -14,6 +15,64 @@ type Messages struct {
...
+	return slices.ContainsFunc(ms, func(m Messages) bool {
+		return m.Language == targetLang
+	})
+}
+
+// LanguageIndex returns index of Messages with the given language. If not found, returns -1.
+func (ms MessagesSlice) LanguageIndex(targetLang language.Tag) int {
```
janishorsts, line 28: Always try to use single-word variable names, and add two or three words if they add meaningful context and meaning.

```suggestion
func (ms MessagesSlice) LanguageIndex(lang language.Tag) int {
```

pkg/server/file.go
```diff
@@ -114,24 +114,47 @@ func (t *TranslateServiceServer) UploadTranslationFile(
...
 		if err != nil {
 			return nil, status.Errorf(codes.Internal, "")
 		}
 
-		if err = t.populateTranslatedMessages(ctx, params.serviceID, messages, allMessages); err != nil {
-			return nil, err
+		originalMessages := messages
+
+		// Replace messages from repo with the new ones, or append if not found.
+		switch idx := all.LanguageIndex(messages.Language); idx {
```
janishorsts, line 129: why not `.Replace()`?

pkg/server/message.go
```diff
@@ -220,32 +220,41 @@ func (t *TranslateServiceServer) UpdateMessages(
...
-		params.serviceID,
-		repo.LoadMessagesOpts{FilterLanguages: []language.Tag{}})
+	all, err := t.repo.LoadMessages(ctx, params.serviceID, repo.LoadMessagesOpts{})
 	if err != nil {
 		return nil, status.Errorf(codes.Internal, "")
 	}
 
-	if len(msgs) == 0 || !langExists(msgs, params.messages.Language) {
-		return nil, status.Errorf(codes.NotFound, "messages not found for language: '%s'", params.messages.Language)
+	switch idx := all.LanguageIndex(params.messages.Language); idx {
```
janishorsts, line 228: why not `.Replace()`?

pkg/server/message.go
```diff
@@ -254,69 +263,150 @@ func (t *TranslateServiceServer) UpdateMessages(
...
+	{ Language: en, Original: true, Messages: [ { ID: "1", Message: "Hello", Status: Translated  } ], ...
+	{ Language: fr, Messages: [ { ID: "1", Message: "Bonjour", Status: Translated  } ], ... ] }
+	{ Language: de, Messages: [ { ID: "1", Message: "Hallo", Status: Translated  } ], ... ]
+
+	Result:
+	{ Language: en, Original: true, Messages: [ { ID: "1", Message: "Hello", Status: Translated  } ], ...
+	{ Language: fr, Messages: [ { ID: "1", Message: "Bonjour", Status: Untranslated  }, ... ] }
+	{ Language: de, Messages: [ { ID: "1", Message: "Hallo", Status: Untranslated  } ], ... ]
+*/
+func (t *TranslateServiceServer) alterTranslations(
```
janishorsts, line 283: Let's add a method to `model.MessagesSlice` instead.

pkg/server/file.go
VladislavsPerkanuks: Wanted to delete that method.

The method's behavior can be confusing because it combines two operations: replacement and appending.

pkg/server/message.go
```diff
@@ -254,69 +263,150 @@ func (t *TranslateServiceServer) UpdateMessages(
...
+getUntranslatedIDs returns a list of message IDs that have been altered e.g.
+ 1. The message.message has been changed
+ 2. The message with new message.ID has been added
+*/
+func getUntranslatedIDs(old, new *model.Messages) []string {
+	lookup := make(map[string]string, len(old.Messages))
+
+	for _, msg := range old.Messages {
+		lookup[msg.ID] = msg.Message
+	}
```
janishorsts, line 401: Can we create a generic helper for a fast check of the list containing a value? Very common functionality.

Something like...

```go
type ContainsFn = func(T comparable) bool

func Contains[T comparable, S any](s []S, f func(S) T) ContainsFn {
    // prepare data for fast lookup

    return func(v T) bool {
        
    }
}
```

pkg/model/message.go
VladislavsPerkanuks: deleted as unused

pkg/server/message.go
VladislavsPerkanuks: We should be able to, but in separate PR, as this one is getting quite large

pkg/model/message.go
```diff
@@ -14,6 +15,83 @@ type Messages struct {
...
+	{ Language: fr, Messages: [ { ID: "1", Message: "Bonjour", Status: Translated  } ], ... ] }
+	{ Language: de, Messages: [ { ID: "1", Message: "Hallo", Status: Translated  } ], ... ]
+
+	Result:
+	{ Language: en, Original: true, Messages: [ { ID: "1", Message: "Hello", Status: Translated  } ], ...
+	{ Language: fr, Messages: [ { ID: "1", Message: "Bonjour", Status: Untranslated  }, ... ] }
+	{ Language: de, Messages: [ { ID: "1", Message: "Hallo", Status: Untranslated  } ], ... ]
+*/
+func (ms MessagesSlice) AlterTranslations(untranslatedIDs []string) {
+	if len(untranslatedIDs) == 0 || len(ms) == 1 {
```
janishorsts, line 75: What is use case for `len(ms) == 1`?

pkg/model/message.go
```diff
@@ -14,6 +15,83 @@ type Messages struct {
...
+	{ Language: en, Original: true, Messages: [ { ID: "1", Message: "Hello", Status: Translated  } ], ...
+	{ Language: fr, Messages: [ { ID: "1", Message: "Bonjour", Status: Translated  } ], ... ] }
+	{ Language: de, Messages: [ { ID: "1", Message: "Hallo", Status: Translated  } ], ... ]
+
+	Result:
+	{ Language: en, Original: true, Messages: [ { ID: "1", Message: "Hello", Status: Translated  } ], ...
+	{ Language: fr, Messages: [ { ID: "1", Message: "Bonjour", Status: Untranslated  }, ... ] }
+	{ Language: de, Messages: [ { ID: "1", Message: "Hallo", Status: Untranslated  } ], ... ]
+*/
+func (ms MessagesSlice) AlterTranslations(untranslatedIDs []string) {
```
janishorsts, line 74: What principle do you apply when choosing what functions to implement in the `model` package? `AlterTranslations` is here, but `populateTranslations` and `getUntranslatedIDs` are in the `server` package. What are their scope and intent?

pkg/model/message.go
VladislavsPerkanuks: Nothing to alter, e.g.
Slice contains only one element, allegedly original -> original files should remain untouched.

But a more straightforward check could be implemented

pkg/model/message.go
VladislavsPerkanuks: Did not want to make extra changes in this PR, as it is already quite large.

But conceptually `populateTranslations` should go to `model` pkg with the receiver `MessagesSlice`. It seems a good fit, by looking at function parameters.

`getUntranslatedIDs` I guess should remain in `server` pkg. As it's scope is limited.

janishorsts: It's worth to make an extra effort. The basic principle we can apply is to move everything to the `model` package that does data transformation - single responsibility.

Also, `AlterTranslations` name is too abstract and does not tell what it does.

```go
// in the handler
all.MarkUntranslated(original.FindUntranslatedIDs(previous))
```

pkg/model/message.go
janishorsts: By reading the code, the reader does not know it. Add the moment?

pkg/server/message.go
```diff
@@ -254,69 +262,54 @@ func (t *TranslateServiceServer) UpdateMessages(
 
 // helpers
 
-// populateTranslatedMessages adds any missing messages from the original messages to all translated messages.
-func (t *TranslateServiceServer) populateTranslatedMessages(
+// fuzzyTranslate fuzzy translates any untranslated messages,
+// returns messagesSlice containing refreshed translations.
+func (t *TranslateServiceServer) fuzzyTranslate(
```
VladislavsPerkanuks, line 267: I am thinking about removing this helper and moving its logic to `fuzzy` pkg, but once again it requires many changes...

It would be more appropriate that Translate takes `[]Messages`, and changes in place all values, instead of returning new `Messages`

Something like
```go
Translate(ctx context.Context, messages ...model.Messages) error
// or
Translate(ctx context.Context, messages model.MessagesSlice) error
```

pkg/model/message.go
```diff
@@ -14,6 +16,177 @@ type Messages struct {
...
+					origMsg.Status = MessageStatusUntranslated
+					ms[i].Messages = append(ms[i].Messages, origMsg)
+				}
+			}
+		}
+
+		go populate(i)
+	}
+
+	wg.Wait()
```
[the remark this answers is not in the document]
VladislavsPerkanuks: Changed.
If my calculations are correct, the space complexity should be **O(1)**. Although the time complexity is high, the utilization of goroutines should speed things up.

pkg/model/message.go
```diff
@@ -14,6 +16,186 @@ type Messages struct {
...
+			Original: true,
+			Messages: [ { ID: "1", Message: "Hello" }, { ID: "2", Message: "World" } ],
+		},
+		{
+			Language: fr,
+			Original: false,
+			Messages: [ { ID: "1", Message: "Bonjour" }, { ID: "2", Message: "World", Status: Untranslated } ],
+		},
+*/
+func (ms MessagesSlice) PopulateTranslations() {
```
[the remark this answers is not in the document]
VladislavsPerkanuks: @janishorsts your call if we should get rid of goroutines or not

pkg/server/file.go
```diff
@@ -114,24 +114,40 @@ func (t *TranslateServiceServer) UploadTranslationFile(
...
+		all, err = t.repo.LoadMessages(ctx, params.serviceID, repo.LoadMessagesOpts{})
 		if err != nil {
 			return nil, status.Errorf(codes.Internal, "")
 		}
 
-		if err = t.populateTranslatedMessages(ctx, params.serviceID, messages, allMessages); err != nil {
-			return nil, err
+		all.Replace(*messages)
+
+		originalMessages := all[all.OriginalIndex()]
```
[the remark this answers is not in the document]
janishorsts: Please fix it

pkg/server/message.go
janishorsts: Scope creep, leave it for another PR. We need to finish this PR. It's taking too long.

pkg/model/message.go
```diff
@@ -14,6 +16,186 @@ type Messages struct {
...
+/*
+FindChangedMessageIDs returns a list of message IDs that have been altered in the new Messages e.g.
+ 1. The message.message has been changed
+ 2. The message with new ID has been added.
+*/
+func (m *Messages) FindChangedMessageIDs(new *Messages) []string {
+	lookup := make(map[string]string, len(m.Messages))
+
+	for _, msg := range m.Messages {
+		lookup[msg.ID] = msg.Message
```
janishorsts, line 28: Storing the index as a value would be faster.

janishorsts (approved):

PR #162, opened 2023-09-14
Repo closer

VladislavsPerkanuks:
Apparently, the problem was not the not closing DB...

When running tests locally, instead of earthly, `cmd/client` and `cmd/service,` tries to acquire a lock on the `TRANSLATE_DB_BADGERDB_PATH` at the same time. On earthly this env variable is empty, which leads to use in memory storage.

Decided to leave `TRANSLATE_DB_BADGERDB_PATH` empty for tests, as it should suffice.  For running the service and trying to upload big files e.g. `Superset`, one should change this env some dir. 

Also left Close function, at least in badgerDB as it is crucial.

### DIFF
--- a/cmd/translate/service/service.go
+++ b/cmd/translate/service/service.go
@@ -79,8 +79,8 @@ var rootCmd = &cobra.Command{
 		}
 
 		defer func() {
-			if repoCloseErr := repo.Close(); repoCloseErr != nil {
-				log.Printf("close repo: %v", repoCloseErr)
+			if closeErr := repo.Close(); closeErr != nil {
+				log.Printf("close repo: %v", closeErr)
 			}
 		}()
 
--- a/cmd/translate/service/service.go
+++ b/cmd/translate/service/service.go
@@ -78,6 +78,12 @@ var rootCmd = &cobra.Command{
 			log.Panicf("create new repo: %v", err)
 		}
 
+		defer func() {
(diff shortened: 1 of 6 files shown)


pkg/repo/repo.go
```diff
@@ -32,4 +32,6 @@ type MessagesRepo interface {
 type Repo interface {
 	ServicesRepo
 	MessagesRepo
+
+	Close() error
```
VladislavsPerkanuks, line 36: **Quick question:**
Are there any advantages of using `io.Closer,` instead of `Close() func` ?

janishorsts: As you know, @VladislavsPerkanuks , golang has implicit interfaces. In general, both ways are fine. However, if an implementation conforms to an interface, it does not meet they have the same intent. As a rule of thumb, use an interface if it matches the intent.

cmd/translate/service/service.go
```diff
@@ -78,6 +78,12 @@ var rootCmd = &cobra.Command{
 			log.Panicf("create new repo: %v", err)
 		}
 
+		defer func() {
+			if repoCloseErr := repo.Close(); repoCloseErr != nil {
```
janishorsts, line 82: keep it simple :)
```suggestion
			if closeErr := repo.Close(); closeErr != nil {
```

janishorsts (approved):

PR #163, opened 2023-09-15
157 Wrong Status For Original Messages

VladislavsPerkanuks:
Status for the message is derived from the `original` and `fuzzy` flag. Explained in here: [diff](https://github.com/translate-agent/translate/pull/163/files#diff-5c855577a14bc1fb4afddc6d09d110989e0a272eb2969bb987effbb965e67f24R14-R20)


kaaposc (approved):

pkg/convert/convert.go
```diff
@@ -32,3 +36,24 @@ func removeEnclosingBrackets(message string) string {
 	replacer := strings.NewReplacer("{", "", "}", "")
 	return replacer.Replace(message)
 }
+
+/*
+getStatus returns the status of a message based on the original and fuzzy flags and the message string.
+
+  - Scenario 1: is original -> translated
+  - Scenario 2: not original and msg is not empty -> translated
```
janishorsts, line 44: This is not true, we cannot rely on the message to determine its status.

pkg/convert/go.go
```diff
@@ -70,15 +70,26 @@ func messagesFromPipeline(m pipeline.Messages, original bool) model.Messages {
 		Original: original,
 	}
 
+	getMessage := func(m pipeline.Message) string { return m.Translation.Msg }
```
VladislavsPerkanuks, line 73: When not original, we should extract from the translation not the message.

pkg/convert/xliff12.go
```diff
@@ -64,19 +64,24 @@ func FromXliff12(data []byte, original bool) (model.Messages, error) {
 	}
 
 	getMessage := func(t transUnit) string { return t.Source }
+	status := model.MessageStatusTranslated
 
 	// Check if a target language is set
 	if !messages.Original {
```
janishorsts, line 70: Can you please replace the negation with the following? It should improve readability.

```suggestion
	if messages.Original {
```

pkg/convert/xliff2.go
```diff
@@ -53,11 +53,13 @@ func FromXliff2(data []byte, original bool) (model.Messages, error) {
 	}
 
 	getMessage := func(u unit) string { return u.Source }
+	status := model.MessageStatusTranslated
 
 	// Check if a target language is set
 	if !messages.Original {
```
janishorsts, line 59: The same here - replace the negation and update the flow.

PR #164, opened 2023-09-21
MessageFormat parse panic

DarjaLeonova:


kaaposc (approved):

pkg/messageformat/lex.go
```diff
@@ -211,12 +211,13 @@ func lexExpr(l *lexer) stateFn {
 		}
 	case '$':
 		return lexVariable(l)
+
 	case ':':
 		return lexFunction(l)
 	case '+':
-		return lexOpeningFunction(l)
+		return processFunction(l, v)
```
janishorsts, line 218: No need for `processFunction`, it makes it harder to read the code, and slower, too.

One option is ...

```suggestion
		if l.exprDepth > 1 {
			return lexOpeningFunction(l)
		}
		
		l.backup()
		
		return lexText(l)
```

pkg/messageformat/lex.go
```diff
@@ -370,3 +371,18 @@ func isNameChar(v rune) bool {
...
+			return lexOpeningFunction(l)
+		case '-':
+			return lexClosingFunction(l)
+		}
+	}
+
+	l.backup()
+
+	return lexText(l)
+}
```
janishorsts, line 388: ... or pass the lex function

```suggestion
func processFunction(l *lexer, f ...) stateFn {
	if l.exprDepth > 1 {
		f(l)
	}

	l.backup()

	return lexText(l)
}
```

DarjaLeonova:
@janishorsts looks like we have the same problem with variable ($) and function (:) chars. 
We know that opening and closing functions are always depthLevel > 1 . {(1)expression block {(2)+button}{-button}(2) }(1)
but variables can be written as "match {$count :number} (depth level == 1)" ... 
So the issue will be when the text block contains $ or :.
Should i fix it in the same PR or create new one ?

janishorsts:
@DarjaLeonova please fix it in this PR, the change is small. No need for micromanagement.

VladislavsPerkanuks (approved):
Seems good

janishorsts (approved):

PR #167, opened 2023-09-26
Make translated default status in message proto

DarjaLeonova:


VladislavsPerkanuks (approved):
Changes in the switch statement were unnecessary :)

Otherwise okay: it seems it does not break any functionality, which is a good thing

DarjaLeonova:
> Changes in the switch statement were unnecessary :)
> 
> Otherwise okay: it seems it does not break any functionality, which is a good thing

DarjaLeonova:
Yeas, I know. I did that to match status order.

PR #175, opened 2023-10-10
172-bugfix-list-translations-doesnt-contain-changes-from-update-translation

klavinsjanis:


pkg/server/translation.go
```diff
@@ -249,12 +249,11 @@ func (t *TranslateServiceServer) UpdateTranslation(
 		}
 
 		updatedTranslations = all
-
 	}
 
 	// Update messages for all translations
 	for i := range updatedTranslations {
-		err = t.repo.SaveTranslation(ctx, params.serviceID, &all[i])
+		err = t.repo.SaveTranslation(ctx, params.serviceID, &updatedTranslations[i])
```
janishorsts, line 256: Having two instances of the same data will cause issues down the road. Is it possible to get rid off updatedTranslations and keep only one `all`?

pkg/server/file.go
```diff
@@ -114,39 +114,42 @@ func (t *TranslateServiceServer) UploadTranslationFile(
...
+		all.MarkUntranslated(oldOriginal.FindChangedMessageIDs(translation))
+		// Replace original translation with new one.
+		all.Replace(*translation)
+		// Add missing messages for all translations.
+		if params.populateTranslations {
+			all.PopulateTranslations()
+		}
 
-			updatedTranslations = all
+		if err := t.fuzzyTranslate(ctx, all); err != nil {
```
VladislavsPerkanuks, line 145: linter will fail on this one when #174 gets merged, as it is a shadow declaration

I suggest changing the base for this PR to the same branch as #174 to avoid future problems

VladislavsPerkanuks (approved):

cmd/translate/main_grpc_integration_test.go
```diff
@@ -683,6 +683,42 @@ func Test_UpdateTranslation_gRPC(t *testing.T) {
...
 	}
 }
+
+// matchingTranslationExistsInService checks incoming translation is equal to translation
+// with same language returned from listTranslations.
+func matchingTranslationExistsInService(
+	ctx context.Context,
+	t *testing.T,
+	serviceID string,
+	t1 *translatev1.Translation,
```
janishorsts, line 700: Please avoid the same short names with different meanings: t - *testing.T, t2 - translatev1.Translation. Do full name.

```suggestion
	translation *translatev1.Translation,
```

janishorsts (approved):

PR #177, opened 2023-10-12
Escape curly braces in other formats

DarjaLeonova:


kaaposc (approved):

pkg/convert/go_test.go
```diff
@@ -127,10 +107,92 @@ func TestFromGo(t *testing.T) {
...
+					{
+						ID:          "2",
+						Message:     "{translatedMessage2}",
+						Description: "description2",
+						Positions:   []string{"src/config.go:20"},
+						Status:      model.MessageStatusFuzzy,
+					},
+					{
+						ID:          "3",
+						Message:     "",
```
klavinsjanis, line 131: FromGo "message": "message3" to our message structure results In Message being empty?

pkg/convert/json_nglocalize_test.go
```diff
@@ -48,6 +48,33 @@ func Test_FromNgLocalize(t *testing.T) {
 				},
 			},
 		},
+		{
+			name: "Message with placeholder",
```
VladislavsPerkanuks, line 52: not sure if in nglocalize in json format, placeholders are in curly braces

pkg/convert/json_ngxtranslate_test.go
```diff
@@ -34,6 +34,20 @@ func Test_FromNgxTranslate(t *testing.T) {
 				},
 			},
 		},
+		{
+			name:  "Message with placeholder",
```
VladislavsPerkanuks, line 38: same, not sure if text in curly braces are considered placeholders

Maybe just:
```suggestion
			name:  "Message with special chars",
```

VladislavsPerkanuks (commented):
I see that every format already had escaped special chars, interesting

pkg/convert/json_ngxtranslate_test.go
klavinsjanis: I think `{world}` could be described as a parameter placeholder.

pkg/convert/arb_test.go
```diff
@@ -21,47 +21,6 @@ func Test_FromArb(t *testing.T) {
...
-						Status:  model.MessageStatusTranslated,
-					},
-					{
-						ID:      "farewell",
-						Message: "{Goodbye friend}",
-						Status:  model.MessageStatusTranslated,
-					},
-				},
-			},
-		},
```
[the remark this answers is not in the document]
DarjaLeonova: We already have exact same test with name "Message with placeholder".

pkg/convert/xliff12_test.go
```diff
@@ -87,9 +88,10 @@ func Test_FromXliff12(t *testing.T) {
 	)
 
 	tests := []struct {
-		name     string
-		expected *model.Translation
-		data     []byte
+		name                string
+		expected            *model.Translation
+		data                []byte
+		containsSpecialChar bool
```
klavinsjanis, line 94: containsParameterPlaceholder

pkg/convert/xliff2_test.go
```diff
@@ -109,6 +112,33 @@ func Test_FromXliff2(t *testing.T) {
 			data:     randXliff2(nonOriginalTranslation),
 			expected: nonOriginalTranslation,
 		},
+		{
+			name:                "Message with special chars",
```
klavinsjanis, line 116: `containsSpecialChar: true` makes the test less understandable, is there any way to remove it?

pkg/convert/json_ngxtranslate_test.go
```diff
@@ -34,6 +34,20 @@ func Test_FromNgxTranslate(t *testing.T) {
 				},
 			},
 		},
+		{
+			name:  "Message with special chars",
```
klavinsjanis, line 38: "Message with placeholder parameter" goes for other tests as well.

pkg/convert/xliff2_test.go
DarjaLeonova: about the name of the test, a Message with special chars also should work fine, cuz it should escape also if there is only a single ("{"or "}" or "|" or "\\") special char inside. (escapeSpecialChars func in convert.go)
can you please explain how containsSpecialChar flag makes it less understandable?  :)

klavinsjanis: well if it escapes those as well then I rest my case :D

pkg/convert/json_nglocalize_test.go
```diff
@@ -48,6 +48,33 @@ func Test_FromNgLocalize(t *testing.T) {
 				},
 			},
 		},
+		{
+			name: "Message with special chars",
+			input: []byte(`
+      {
+        "locale": "fr",
+        "translations": {
+          "Hello": "Bonjour {le monde}",
```
klavinsjanis, line 57: Perhaps it would be good to find real-world examples to supplement the test with more cases.

VladislavsPerkanuks (approved):

pkg/convert/xliff12_test.go
```diff
@@ -101,6 +103,33 @@ func Test_FromXliff12(t *testing.T) {
 			data:     randXliff12(nonOriginalTranslation),
 			expected: nonOriginalTranslation,
 		},
+		{
```
klavinsjanis, line 106: Maybe you could supplement test with more cases?
Example: 
Message: `Order #{Id} has been canceled for {ClientName}`,

pkg/convert/go_test.go
```diff
@@ -127,10 +107,92 @@ func TestFromGo(t *testing.T) {
...
+		},
+		{
+			name: "Message with special chars",
+			input: []byte(`
+	{
+		"language": "en",
+		"messages": [
+			{
+				"id": "1",
+				"meaning": "description1",
```
klavinsjanis, line 146: Maybe you could supplement test with more cases?
Example:
Message: Order #{Id} has been canceled for {ClientName},

pkg/convert/xliff12_test.go
```diff
@@ -101,6 +103,33 @@ func Test_FromXliff12(t *testing.T) {
...
+					},
+				},
+			),
+			expected: &model.Translation{
+				Original: false,
+				Language: language.English,
+				Messages: []model.Message{
+					{
+						ID:      "order canceled",
+						Message: `{Order #\{Id\} has been canceled for \{ClientName\} \| \\}`,
```
klavinsjanis, line 127: Shouldn't this be -> `{Order #\{Id\} has been canceled for \{ClientName\} \| \\\\}` ?

pkg/convert/arb_test.go
```diff
@@ -22,14 +22,14 @@ func Test_FromArb(t *testing.T) {
 	}{
```
klavinsjanis, line 22: What about Test_ToArb, Test_ToGo etc. tests?

pkg/convert/xliff12_test.go
DarjaLeonova: The expected Message is correct. since string "\\" is equal to single `\` in backticks. 
I will also wrap the input message in backticks for readability.

pkg/convert/arb.go
```diff
@@ -135,7 +135,7 @@ func ToArb(translation model.Translation) ([]byte, error) {
 	dst["@@locale"] = translation.Language
 
 	for _, msg := range translation.Messages {
-		dst[msg.ID] = removeEnclosingBrackets(msg.Message)
+		dst[msg.ID] = removeEscapeSpecialChars(removeEnclosingBrackets(msg.Message))
```
janishorsts, line 138: Would parsing the MF2 and using the text node work better? This way we get the real value (unserialized).

pkg/convert/arb.go
```diff
@@ -131,11 +131,18 @@ func FromArb(data []byte, original bool) (model.Translation, error) {
...
 	// "und" (Undetermined) language.Tag is also valid BCP47 tag.
 	dst["@@locale"] = translation.Language
 
 	for _, msg := range translation.Messages {
-		dst[msg.ID] = removeEnclosingBrackets(msg.Message)
+		var err error
+
+		dst[msg.ID], err = getMsg(msg.Message)
+		if err != nil {
+			return nil, fmt.Errorf("parse NodeText %w", err)
```
janishorsts, line 143:
```suggestion
			return nil, fmt.Errorf("get message value: %w", err)
```

pkg/convert/convert.go
```diff
@@ -29,8 +33,20 @@ func escapeSpecialChars(message string) string {
...
+		return "", fmt.Errorf("parse message: %w", err)
+	}
+
+	for _, node := range nodes {
+		nodeTxt, ok := node.(messageformat.NodeText)
+		if !ok {
+			return "", errors.New("convert node to messageformat.NodeText")
+		} else {
+			return nodeTxt.Text, nil
+		}
```
janishorsts, line 48: if you have `if... else`, do not negate the expression, it reduces readability. Do not use the `else` block if `if` returns.

```suggestion
		if ok {
			
			return nodeTxt.Text, nil
		}
		
		return "", errors.New("convert node to messageformat.NodeText")
```

janishorsts (approved):

klavinsjanis (approved):

PR #188, opened 2023-10-27
180 3.Extract Placeholders From Source Text

VladislavsPerkanuks:
Initial implementation to extract placeholders from source text using regex.

Some screenshots when uploading superset pot file
image
image
image


pkg/convert/pot_test.go
```diff
@@ -1148,8 +1158,8 @@ when * {Il y a {$count} pommes.}
 						ID:       "There is %d apple.",
 						PluralID: "There are %d apples.",
 						Message: `match {$count :number}
-when 1 {There is {$count} apple.}
-when * {There are {$count} apples.}
+when 1 {There is {:Placeholder format=printf type=int} apple.}
+when * {There are {:Placeholder format=printf type=int} apples.}
```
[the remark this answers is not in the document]
VladislavsPerkanuks: we dont know what we are matching.

two things:

1. if text has two placeholders, e.g: `blah %d blah %d`
that would be incorrect to convert it to: `{blah {$count :Placeholder ...} blah {$count :Placeholder}}`
2. if text has other type of placeholder with variable name e,g.; `blah {num1}`
that would be incorrect to lose original variable name: e.g. `blah {$count :Placeholder ...}`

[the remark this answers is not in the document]
janishorsts: Interesting.

pkg/convert/pot.go
```diff
@@ -76,12 +76,14 @@ func FromPot(b []byte, original bool) (model.Translation, error) {
...
-	convert := func(v pot.MessageNode) string { return convertToMessageFormatSingular(singularValue(v)) }
+	convert := func(v pot.MessageNode) string {
+		return messageformat.ToMessageFormat2(singularValue(v))
+	}
 
 	if po.Header.PluralForms.NPlurals == pluralCountLimit {
 		convert = func(v pot.MessageNode) string {
 			if v.MsgIDPlural == "" {
-				return convertToMessageFormatSingular(singularValue(v))
+				return messageformat.ToMessageFormat2(singularValue(v))
```
klavinsjanis, line 86: Interesting function name. Are you converting message format to message format 2?

VladislavsPerkanuks: source(extracted) string to message format

klavinsjanis: To make the intent of the function more obvious, perhaps - mf2.MessageFromString()?

pkg/messageformat/extract.go
```diff
@@ -0,0 +1,193 @@
...
+		nodeExprF: createNodeExpr("bracketVar"),
+	},
+	"emptyBracket": { // hello {}
+		re:        regexp.MustCompile(`\{\}`),
+		nodeExprF: createNodeExpr("emptyBracket"),
+	},
+}
+
+// messageToAST converts the given message to an MF2 AST.
+func (pf *placeholderFormat) messageToAST(message string) AST {
```
klavinsjanis, line 65: why not pass in *placeholderFormat as an argument?

pkg/convert/pot.go
```diff
@@ -337,14 +345,16 @@ func writeTags(b *bytes.Buffer, message model.Message) error {
 }
 
 // convertPluralsToMessageString converts a slice of strings to MessageFormat plural form.
-func convertPluralsToMessageString(plurals []string) string {
+func convertPluralsToMessageString(plurals []string) (string, error) {
```
klavinsjanis, line 348: Maybe could simplify the function name e.g. messageFromPlurals()?

pkg/messageformat/extract.go
VladislavsPerkanuks: just thought, that it belongs as a method for struct, more than an standalone function 🤷

pkg/convert/pot.go
VladislavsPerkanuks: image

pkg/convert/pot_test.go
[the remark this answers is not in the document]
janishorsts: How does the translation library (e.g. in python) handle the given case - choose a singular or plural sentence? Based on `%d`, right?

What is the intent behind `Plural-Forms: nplurals=2; plural=(n != 1);\n` in the POT? What does the code look like in Python to handle it?

pkg/convert/pot.go
janishorsts: There are common interfaces we can use to convey serialisation - MarshalText & UnmarshalText?

pkg/messageformat/extract.go
```diff
@@ -0,0 +1,193 @@
+package messageformat
+
+import (
+	"fmt"
+	"regexp"
+	"strings"
+)
+
+// specialChars is a list of characters that need to be escaped in the node.Text,
```
janishorsts, line 9: Please add links to MF2 syntax.

pkg/messageformat/extract.go
janishorsts: @klavinsjanis , there are two benefits of using pointer receivers:
* make changes to the receiver instance (as you said)
* avoid copying large instances by value. E.g. if the receiver instance uses 1KB of memory, passing by value is very expensive, passing by pointer makes it cheaper.

pkg/messageformat/extract.go
```diff
@@ -0,0 +1,193 @@
...
+		return NodeExpr{Function: NodeFunction{Name: "Placeholder", Options: options}}
+	}
+}
+
+// escapeSpecialChars escapes special characters in the given text.
+func escapeSpecialChars(text string) string {
+	var sb strings.Builder
+
+	for _, c := range text {
+		if _, ok := specialChars[c]; ok {
```
janishorsts, line 154:
```suggestion
		switch c {
			case '{', '}',....:
				sb.WriteRune('\\')
		}
```

pkg/messageformat/extract.go
klavinsjanis: I know about pointers, but just was wondering why method is used instead of function.

Issue #181, opened 2023-10-23
FromPOT: Some messages with plurals are merging


VladislavsPerkanuks:
Long story short:

In .pot files which contain extracted texts, there is no `Plural-Forms`header (this header is only present in .po files e.g. translated ones), as a result, there is unexpected behaviour when importing

test case:
```
msgid ""
msgstr ""
"Language: en\n"

#: superset-frontend/src/filters/components/GroupBy/GroupByFilterPlugin.tsx:87
#: superset-frontend/src/filters/components/Select/SelectFilterPlugin.tsx:226
#: superset-frontend/src/filters/components/TimeColumn/TimeColumnFilterPlugin.tsx:86
#: superset-frontend/src/filters/components/TimeGrain/TimeGrainFilterPlugin.tsx:96
#, python-format
msgid "%s message"
msgid_plural "%s messages"
msgstr[0] ""
msgstr[1] ""

#: superset/views/core.py:385
#, python-format
msgid "message2"
msgstr ""
```

When calling `FromPot()`
<img width="506" alt="image" src="https://github.com/translate-agent/translate/assets/69954124/b1c5f1a9-bf82-4e44-87a0-1f28e0a7a0cc">

expected length =  2
actual length = 1

PR #198, opened 2023-11-02
181-optional PluralForm header

DarjaLeonova:
https://github.com/translate-agent/translate/issues/181

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -20,6 +20,7 @@ linters:
    - goconst
    - gochecknoglobals
    - gochecknoinits
+   - gocyclo
    - godox
    - goerr113
    - gofmt
--- a/pkg/convert/pot.go
+++ b/pkg/convert/pot.go
@@ -80,7 +80,7 @@ func FromPot(b []byte, original bool) (model.Translation, error) {
 
 	if po.Header.PluralForms.NPlurals == pluralCountLimit {
 		convert = func(v pot.MessageNode) string {
-			if v.MsgIDPlural == "" || allEmpty(pluralValue(v)) {
+			if v.MsgIDPlural == "" {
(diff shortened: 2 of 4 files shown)


pkg/convert/pot.go
```diff
@@ -80,7 +80,7 @@ func FromPot(b []byte, original bool) (model.Translation, error) {
 
 	if po.Header.PluralForms.NPlurals == pluralCountLimit {
 		convert = func(v pot.MessageNode) string {
-			if v.MsgIDPlural == "" {
+			if v.MsgIDPlural == "" || allEmpty(pluralValue(v)) {
```
janishorsts, line 83: What is `po` instance value for the case described by @VladislavsPerkanuks ?

mvilks (approved):

klavinsjanis (approved):

VladislavsPerkanuks (approved):

janishorsts (approved):

.golangci.yml
```diff
@@ -20,6 +20,7 @@ linters:
    - goconst
    - gochecknoglobals
    - gochecknoinits
+   - gocyclo
```
janishorsts, line 23: What is the reason behind disabling the `gocyclo` linter?

DarjaLeonova: "cyclop" and "gocyclo" are similar linters except that "cyclop" also scans packages (gocycle scnas only functions).

From their github readme:
Gocyclo calculates [cyclomatic complexities](https://en.wikipedia.org/wiki/Cyclomatic_complexity) of functions in Go source code. "Source: https://github.com/fzipp/gocyclo"
Cyclop calculates cyclomatic complexities of functions or packages in Go source code. "Source: https://github.com/bkielbasa/cyclop"

So we decided to disable gocyclo.

Issue #182, opened 2023-10-23
ToPot: Plural-Forms header should be present only once, and only when original = false


VladislavsPerkanuks:
```go
input: model.Translation{
	Language: language.French,
        Original: false,
	Messages: []model.Message{
		{
			ID:          "There is %d apple.",
			PluralID:    "There are %d apples.",
			Message:     "match {$count :number}\nwhen 1 {Il y a {$count} pomme.}\nwhen * {Il y a {$count} pommes.}",
			Description: "apple counts",
			Status:      model.MessageStatusFuzzy,
		},
		{
			ID:          "There is %d apple.",
			PluralID:    "There are %d apples.",
			Message:     "match {$count :number}\nwhen 1 {Il y a {$count} pomme.}\nwhen * {Il y a {$count} pommes.}",
			Description: "apple counts",
			Status:      model.MessageStatusFuzzy,
		},
	},
},
```

Expected:
```go
msgid ""
msgstr ""
"Language: en\n"
"Plural-Forms: nplurals=2; plural=(n != 1);\n" 

#. apple counts
#, fuzzy
msgid "There is %d apple."
msgid_plural "There are %d apples."
msgstr[0] "Il y a %d pomme."
msgstr[1] "Il y a %d pommes."

#. apple counts
#, fuzzy
msgid "There is %d apple."
msgid_plural "There are %d apples."
msgstr[0] "Il y a %d pomme."
msgstr[1] "Il y a %d pommes."
```

Actual:
```go
msgid ""
msgstr ""
"Language: en\n"
"Plural-Forms: nplurals=2; plural=(n != 1);\n"  // <---
#. apple counts
#, fuzzy
msgid "There is %d apple."
msgid_plural "There are %d apples."
msgstr[0] "Il y a %d pomme."
msgstr[1] "Il y a %d pommes."

"Plural-Forms: nplurals=2; plural=(n != 1);\n"  // <---
#. apple counts
#, fuzzy
msgid "There is %d apple."
msgid_plural "There are %d apples."
msgstr[0] "Il y a %d pomme."
msgstr[1] "Il y a %d pommes."
```

PR #199, opened 2023-11-03
182- plural header shows only once when have multiple plural strings

DarjaLeonova:
https://github.com/translate-agent/translate/issues/182

### DIFF
--- a/pkg/convert/pot.go
+++ b/pkg/convert/pot.go
@@ -34,6 +34,8 @@ func ToPot(t model.Translation) ([]byte, error) {
 	}
 
 	if !t.Original {
+		// Temporary we support plural forms (one and other).
+		// https://www.gnu.org/software/gettext/manual/html_node/Plural-forms.html
 		if _, err := fmt.Fprintf(&b, "\"Plural-Forms: nplurals=%d; plural=(n != 1);\\n\"\n\n", pluralCountLimit); err != nil {
 			return nil, fmt.Errorf("write Plural-Forms: %w", err)
 		}
--- a/pkg/convert/pot.go
+++ b/pkg/convert/pot.go
@@ -34,8 +34,8 @@ func ToPot(t model.Translation) ([]byte, error) {
 	}
 
 	if !t.Original {
-		if err := writePluralHeader(&b, t); err != nil {
-			return nil, fmt.Errorf("write plural form header: %w", err)
+		if _, err := fmt.Fprintf(&b, "\"Plural-Forms: nplurals=%d; plural=(n != 1);\\n\"\n\n", pluralCountLimit); err != nil {
+			return nil, fmt.Errorf("write Plural-Forms: %w", err)
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/convert/pot_test.go
```diff
@@ -708,6 +793,8 @@ msgstr "Au revoir!"
 			result, err := ToPot(tt.input)
 			require.NoError(t, err)
 
+			t.Logf("actual: %v", string(result))
+			t.Logf("expect: %v", string(tt.expected))
```
VladislavsPerkanuks, line 797: :)
```suggestion
```

pkg/convert/pot_test.go
```diff
@@ -625,6 +628,88 @@ msgstr[0] ""
...
+						Message:     "match {$count :number}\nwhen 1 {Il y a {$count} pomme.}\nwhen * {Il y a {$count} pommes.}",
+						Description: "apple counts",
+						Status:      model.MessageStatusFuzzy,
+					},
+				},
+			},
+			expected: []byte(`msgid ""
+msgstr ""
+"Language: fr\n"
+#. apple counts
```
VladislavsPerkanuks, line 700: missing newline ?
```suggestion
msgstr ""
"Language: fr\n"

#. apple counts
...
```

pkg/convert/pot.go
```diff
@@ -31,6 +33,12 @@ func ToPot(t model.Translation) ([]byte, error) {
 		return nil, fmt.Errorf("write language: %w", err)
 	}
 
+	if !t.Original {
+		if err := writePluralHeader(&b, t); err != nil {
```
VladislavsPerkanuks, line 37: To simplify everything, we can add Plural headers to non-original files, without checking, if they actually have plurals.
e.g.
```suggestion
		fmt.Fprintf(b, "\"Plural-Forms: nplurals=%d; plural=(n != 1);\\n\"\n\n", pluralCountLimit)
```
Somewhere later, we could adjust that header based on language

pkg/convert/pot.go
```diff
@@ -21,6 +21,8 @@ const (
 	MsgStr       poTag = "msgstr"
 )
 
+const pluralCountLimit = 2
```
janishorsts, line 24: In general, keep everything as close as possible! If const is used within a single function, do not declare it at the package level. Declare it inside the function that consumes it.

There are no benefits to declaring it at the package level.

pkg/convert/pot.go
```diff
@@ -40,10 +48,22 @@ func ToPot(t model.Translation) ([]byte, error) {
 	return b.Bytes(), nil
 }
 
+func writePluralHeader(b *bytes.Buffer, t model.Translation) error {
```
janishorsts, line 51: is the plural header the same for the whole translation file? Please add a description and URL to POT documentation to ease understanding.

pkg/convert/pot.go
```diff
@@ -40,10 +48,22 @@ func ToPot(t model.Translation) ([]byte, error) {
...
+	for _, message := range t.Messages {
+		if message.PluralID != "" {
+			if _, err := fmt.Fprintf(b, "\"Plural-Forms: nplurals=%d; plural=(n != 1);\\n\"\n\n", pluralCountLimit); err != nil { //nolint:lll
+				return fmt.Errorf("write plural forms: %w", err)
+			}
+
+			break
```
janishorsts, line 58: `return` conveys the meaning and intent better. If you use `break`, it means there is additional processing but there is none actually.

```suggestion
			return nil
```

pkg/convert/pot.go
DarjaLeonova: A constant pluralCountLimit is used in two different functions. (FromPot and ToPot).

VladislavsPerkanuks (approved):

janishorsts (approved):

PR #204, opened 2023-11-14
203-restrict-creating-multiple-original-translations-per-service

klavinsjanis:


cmd/translate/main_grpc_integration_test.go
```diff
@@ -46,12 +46,23 @@ func randUploadRequest(t *testing.T, serviceID string) *translatev1.UploadTransl
 	return &translatev1.UploadTranslationFileRequest{
 		ServiceId: serviceID,
 		Language:  lang.String(),
-		Original:  gofakeit.Bool(),
 		Data:      data,
 		Schema:    schema,
 	}
 }
 
+func createUpdateTranslationReq(t *testing.T, serviceID string, override *translatev1.Translation,
```
janishorsts, line 54: I think we prefix functions with `create` if it does send a GRPC request. And prefix with `rand` to generate a random entity.

pkg/server/file.go
```diff
@@ -129,6 +132,11 @@ func (t *TranslateServiceServer) UploadTranslationFile(
 		// Original translation is not affected, changes will not affect other translations - update incoming translation.
 		all = model.Translations{*translation}
 	case translation.Original && origIdx != -1:
+		if translation.Language != all[origIdx].Language {
+			return nil, status.Errorf(
+				codes.AlreadyExists, "original translation already exists for service: '%s'", params.serviceID)
```
janishorsts, line 137: `Already exists` applies to the same entity. The better error is `conflict`.

pkg/server/translation.go
```diff
@@ -67,34 +67,30 @@ func (t *TranslateServiceServer) CreateTranslation(
...
 	if err != nil {
 		return nil, status.Errorf(codes.Internal, "")
 	}
 
-	if len(translations) > 0 {
-		return nil, status.Errorf(codes.AlreadyExists,
-			"translation already exist for language: '%s'", params.translation.Language)
+	if all.HasLanguage(params.translation.Language) {
+		return nil,
+			status.Errorf(codes.AlreadyExists, "translation already exists for language: '%s'", params.translation.Language)
```
janishorsts, line 77: conflict

pkg/server/translation.go
```diff
@@ -67,34 +67,30 @@ func (t *TranslateServiceServer) CreateTranslation(
...
-	// noop
-	case false: // Translate messages when translation is not original and original language is known.
-		// TODO: to improve performance should be replaced with CheckTranslationExist db function.
-		loadTranslations, err := t.repo.LoadTranslations(ctx, params.serviceID, repo.LoadTranslationsOpts{})
-		if err != nil {
-			return nil, status.Errorf(codes.Internal, "")
+	case true:
+		if all.OriginalIndex() != -1 {
+			return nil, status.Errorf(
+				codes.AlreadyExists, "original translation already exists for service: '%s'", params.serviceID)
```
janishorsts, line 84: conflict

pkg/server/translation.go
```diff
@@ -229,6 +225,11 @@ func (t *TranslateServiceServer) UpdateTranslation(
 		// Original translation is not affected, changes will not affect other translations - update incoming translation.
 		all = model.Translations{*params.translation}
 	case params.translation.Original && origIdx != -1:
+		if params.translation.Language != all[origIdx].Language {
+			return nil, status.Errorf(
+				codes.AlreadyExists, "original translation already exists for service: '%s'", params.serviceID)
```
janishorsts, line 230: conflict

pkg/server/file.go
klavinsjanis: hmm, couldn't find conflict status code for grpc.

janishorsts: in that case, we can use INVALID_ARGUMENT

pkg/server/translation.go
klavinsjanis: I think this is something else

pkg/server/file.go
```diff
@@ -114,6 +114,9 @@ func (t *TranslateServiceServer) UploadTranslationFile(
 		return nil, status.Errorf(codes.InvalidArgument, err.Error())
 	}
 
+	// Original status for translation is taken from request parameters,
+	translation.Original = params.original
```
VladislavsPerkanuks, line 118: redundant?

`TranslationFromData()` takes params, and creates translation with the original flag set to correct value from params, or file itself if possible,
[code](https://github.com/translate-agent/translate/blob/4fb9df5bde269292ab753dc7cd761e8f4459fb91/pkg/server/convert.go#L43C2-L43C56)

klavinsjanis: I added it there because tests were failing, will look into it.

VladislavsPerkanuks: the only case that could fail could be xliff, because it is the only one that can be determined from source file, afaik

klavinsjanis: Shouldn't we ignore what's in the file? What matters should be parameter passed in through request.

proto/translate/v1/translate.proto
```diff
@@ -49,7 +49,7 @@ message UploadTranslationFileRequest {
   bytes data = 2;
   Schema schema = 3;
   string service_id = 4;
-  bool original = 5;
+  optional bool original = 5;
```
VladislavsPerkanuks, line 52: Hmm, did not know that. 

The last time I checked all proto fields were implicitly optional, but by making it explicitly optional, the field becomes a pointer in the generated file.

Nevertheless, in other languages e.g. cpp, the generated code contains `HasField` method, which can check if the field is set or not, but here we only have one getter, which returns false if not set, and the field value if it is set, so that means there is no way, to distinguish between zero value (false), and set value (false), when we are using getter...

The solution that Jānis used, is to get field omitting getter, but that beats the purpose of getter which is to ensure null safety.

Summarizing all that, got one thing to say: interesting :)

cmd/translate/main_integration_test.go
```diff
@@ -99,3 +99,8 @@ func mustGetFreePort() string {
 
 	return strconv.Itoa(addr.Port)
 }
+
+// ptr returns pointer to the passed in value.
+func ptr[T any](v T) *T {
+	return &v
+}
```
VladislavsPerkanuks, line 106: this is used multiple places, please move to pkg to avoid redeclaring again and again

klavinsjanis: Just a local helper func.
which pkg do you think this should belong to?

VladislavsPerkanuks: `pkg/util` ? or something like that

VladislavsPerkanuks: not local, it is used atleast 3 times in different places, and I forsee that it will be used more
image

cmd/client/cmd/upload.go
```diff
@@ -52,9 +52,15 @@ func newUploadCmd() *cobra.Command {
...
 			}
 
-			original, err := cmd.Flags().GetBool("original")
-			if err != nil {
-				return fmt.Errorf("upload file: get cli parameter 'original': %w", err)
+			var original *bool
+
+			if cmd.Flags().Changed("original") {
+				v, flagErr := cmd.Flags().GetBool("original")
+				if flagErr != nil {
```
janishorsts, line 59:
```suggestion
				v, err := cmd.Flags().GetBool("original")
				if err != nil {
```

pkg/convert/xliff12.go
```diff
@@ -51,7 +51,7 @@ type context struct {
 
 // FromXliff12 converts serialized data from the XML data in the XLIFF 1.2 format into a model.Translation struct.
 // For now original param is ignored.
```
janishorsts, line 53: The comment `// For now original param is ignored.` does not reflect implementation.

pkg/server/convert.go
```diff
@@ -19,28 +19,35 @@ TranslationFromData converts in specific schema serialized data to model.Transla
...
+	)
+
+	if params.original != nil {
+		original = *params.original
+	}
 
 	switch params.schema {
 	case translatev1.Schema_ARB:
-		from = convert.FromArb
+		translation, err = convert.FromArb(params.data, original)
```
janishorsts, line 34: We can keep it as it was before, let `From<std>` determine the `original` from the file if it can.

And simply do once in this function body, right?

```go
  if params.original != nil {
    translation.Original = *params.original
  }
```

pkg/server/convert.go
```diff
@@ -19,28 +19,35 @@ TranslationFromData converts in specific schema serialized data to model.Transla
...
+		translation, err = convert.FromNgLocalize(params.data, original)
 	case translatev1.Schema_JSON_NGX_TRANSLATE:
-		from = convert.FromNgxTranslate
+		translation, err = convert.FromNgxTranslate(params.data, original)
 	case translatev1.Schema_POT:
-		from = convert.FromPot
+		translation, err = convert.FromPot(params.data, original)
 	case translatev1.Schema_XLIFF_2:
-		from = convert.FromXliff2
+		translation, err = convert.FromXliff2(params.data, params.original)
```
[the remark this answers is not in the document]
janishorsts: I believe we don't need to change the func signature at all. Let `From<std>` determine the original from the file, and let the func consumer override it if needed.

pkg/server/convert.go
janishorsts: I can see we passed the original to the `From<std>`. Do we have to? Does it impact the flow and processing of the file?

janishorsts: Ignore my comment. I looked at he code and the `original` affects processing the translation file. 

Do https://github.com/translate-agent/translate/pull/204/files#r1400453299

cmd/client/cmd/upload.go
klavinsjanis: That's there to avoid linter from yelling, I will add nolint then.

pkg/convert/arb.go
```diff
@@ -94,11 +99,11 @@ func FromArb(data []byte, original bool) (model.Translation, error) {
 	}
 
 	status := model.MessageStatusUntranslated
-	if original {
+	if *original {
 		status = model.MessageStatusTranslated
 	}
 
-	translation := model.Translation{Language: lang, Original: original}
+	translation := model.Translation{Language: lang, Original: *original}
```
janishorsts, line 106: simply do

```suggestion
	translation := model.Translation{Language: lang, Original: original != nil && original}
```

pkg/convert/go.go
```diff
@@ -26,14 +26,19 @@ func ToGo(t model.Translation) ([]byte, error) {
...
 		return model.Translation{}, fmt.Errorf("decode JSON to pipeline.Messages: %w", err)
 	}
 
-	return translationFromPipeline(pipelineMsgs, original), nil
+	// if original is not provided default to false.
+	if original == nil {
+		original = ptr(false)
+	}
+
+	return translationFromPipeline(pipelineMsgs, *original), nil
```
janishorsts, line 41:
```suggestion
	return translationFromPipeline(pipelineMsgs, original != nil && original), nil
```

pkg/convert/json_nglocalize.go
```diff
@@ -19,20 +19,25 @@ type ngJSON struct {
...
+	// if original is not provided default to false.
+	if original == nil {
+		original = ptr(false)
+	}
+
 	translation := model.Translation{
 		Language: ng.Language,
 		Messages: make([]model.Message, 0, len(ng.Translations)),
-		Original: original,
+		Original: *original,
```
janishorsts, line 36:
```suggestion
		Original: original != nil && original,
```

janishorsts (approved):

PR #208, opened 2023-11-22
206-fix-convert-to-xliff

klavinsjanis:
To address randomly failing integration tests in:
203-conflict-when-multiple-original-translations

### DIFF
--- a/pkg/convert/xliff12.go
+++ b/pkg/convert/xliff12.go
@@ -96,11 +96,16 @@ func ToXliff12(translation model.Translation) ([]byte, error) {
 	xlf := xliff12{
 		Version: "1.2",
 		File: xliff12File{
-			SourceLanguage: translation.Language,
 			Body: bodyElement{
 				TransUnits: make([]transUnit, 0, len(translation.Messages)),
 			},
 		},
+	}
+
+	if translation.Original {
+		xlf.File.SourceLanguage = translation.Language
+	} else {
+		xlf.File.TargetLanguage = translation.Language
 	}
 
 	for _, msg := range translation.Messages {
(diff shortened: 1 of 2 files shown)


pkg/convert/xliff12.go
```diff
@@ -96,25 +96,37 @@ func ToXliff12(translation model.Translation) ([]byte, error) {
...
 				TransUnits: make([]transUnit, 0, len(translation.Messages)),
 			},
 		},
 	}
 
+	if translation.Original {
+		xlf.File.SourceLanguage = translation.Language
+	} else {
+		xlf.File.TargetLanguage = translation.Language
+	}
```
janishorsts, line 109: When flow is without an else block, the compiled binary code is identical to your flow `if ... else`. I like this better. Your call.

```suggestion
	xlf.File.TargetLanguage = translation.Language
	if translation.Original {
		xlf.File.SourceLanguage = translation.Language
	}
```

klavinsjanis: I am confused, how is that identical?

pkg/convert/xliff2.go
```diff
@@ -104,16 +104,27 @@ func ToXliff2(translation model.Translation) ([]byte, error) {
 		},
 	}
 
+	if translation.Original {
+		xlf.SrcLang = translation.Language
+	} else {
+		xlf.TrgLang = translation.Language
```
janishorsts, line 110: Does xliff12 have only TrgLang or both when it is translated?

* when original, only SrcLang is set
* when not original, SrcLang and TgtLang are set

klavinsjanis: This is just a minor fix, addressing the failing integration tests. The implementation is incomplete for convert to XLIFF. In the current state it's not accurate/incomplete. I have added a ticket:
https://github.com/orgs/translate-agent/projects/1?pane=issue&itemId=45498477

janishorsts (approved):

PR #211, opened 2023-11-24
Tidy PO lexer

VladislavsPerkanuks:
Main reason of this PR

```
// Given example
msgid ""
"line1"
"line"

// Previous behaviour
token == "line1 line2"

// New behaviour
token == "\nline1\nline2"

// or raw (for more visuals)
`
line1
line2
`
```

Also fixed some strange behavior when newline character would be interpreted as an actual newline, not a text
e.g.
```
msgid ""
"line1\n"
"line2\n"

// Previous behaviour
token == "line1 \nline2\n" // not sure but something like that, I hope idea is clear

// New behaviour
token == "\nline1\\n\nline2\\n"

// or raw (for more visuals)
`
line1\n
line2\n
`


```

Also fixed an issue, in a situation, when the header is not defined in TokenType, then the parser tries to parse it as multiline msgid, msgstr e.t.c, and breaks. Either way it should throw an error, but in current implementation it thorws incorrect one

Also some tidy to lexer, to decrease technical debt :)

superset.pot | model.Translation |
---|---|
image |  image


pkg/convert/pot.go
```diff
@@ -344,7 +344,7 @@ func convertPluralsToMessageString(plurals []string) string {
 
 	for i, plural := range plurals {
 		plural = escapeSpecialChars(plural)
-		line := strings.ReplaceAll(strings.TrimSpace(plural), "%d", "{$count}")
```
VladislavsPerkanuks, line 347: We cannot lose information

pkg/pot/parse.go
```diff
@@ -66,9 +66,6 @@ func TokensToPo(tokens []Token) (Po, error) {
 			}
 		}
 
-		replacer := strings.NewReplacer("\\n ", "\n", "\\n", "\n")
-		token.Value = replacer.Replace(token.Value)
```
VladislavsPerkanuks, line 70: no clue what that was for :)

pkg/convert/pot_test.go
```diff
@@ -1005,7 +1005,7 @@ func TestFromPot(t *testing.T) {
 				Messages: []model.Message{
 					{
 						ID:          "Hello",
-						Message:     "{Hello, world!\nvery long string\n}",
+						Message:     "{\nHello, world!\\\\n\nvery long string\\\\n}",
```
janishorsts, line 1008: Why so many escapes?

Escaped `\n` is escaped as string literal `\\n` in golang.
Isn't `\n` escaped as`\\n` in n MF2, too?


```suggestion
						Message:     "{\nHello, world!\\n\nvery long string\\n}",
```

janishorsts: Ok, after some thinking - `\n` should be escaped in `\\n` in MF2, right? Now, to translate it into golang string literal we have to escape both slashes - `\\\\n`.

VladislavsPerkanuks: po has raw text:
`"Hello world\n"` (\n is not newline here, but \\\n)
after converting it to golang string it becomes
`"Hello world\\n"`
as message format tells us that in the text backslashes must be escaped it becomes:
`"Hello world\\\\n"`

VladislavsPerkanuks: in one image
image

pkg/pot/lex.go
```diff
@@ -215,169 +99,140 @@ func Lex(r io.Reader) ([]Token, error) {
...
 
-// hasValidPrefix determines whether the provided line string has a valid prefix
-// by iterating through a collection of valid prefixes.
-func hasValidPrefix(line string) bool {
-	for _, prefix := range validPrefixes {
-		if strings.HasPrefix(line, prefix) {
-			return true
-		}
-	}
+// trimQuotes removes first and last double quotes from the string, if they exist.
```
janishorsts, line 232: No need to create a helper func. Simply use directly `strings.Trim(v, `"`).

PS. There is a side effect that I don't like - quotes are not optional when parsing. They are required.

janishorsts: I mean, we should error if there are no quotes present.

VladislavsPerkanuks: they are required in msgid, msgstr, and in headers
but for example in comments they are not.

Also they have differences

as in 


```
"Language: en-US\n" // prefix: Language: Value: en-US\n"
msgid "some text"   // prefix msgid: Value: "some-text"
#, python-format    // prefix:  # Value: python-format
```

janishorsts: We have to handle it correctly.

VladislavsPerkanuks: reverted to simple strings.Trim()

reason: Too many scenarios to handle.  Also case where file is formatted correctly is more common, than incorrectly (Pareto). So I just go with that, to save time

pkg/pot/lex.go
```diff
@@ -197,9 +82,9 @@ func Lex(r io.Reader) ([]Token, error) {
 			continue
 		}
 
-		token, err := parseLine(line, &tokens)
+		token, err := parseLine(line, tokens)
 		if err != nil {
-			return nil, fmt.Errorf("parse line %d: %w", lineNumber, err)
+			return nil, fmt.Errorf("parse line \"%s\": %w", line, err)
```
janishorsts, line 87: alternatively, easier to read with backquotes

```suggestion
			return nil, fmt.Errorf(`parse line "%s": %w`, line, err)
```

janishorsts (approved):

PR #213, opened 2023-11-29
MF2 Complex Message Support

DarjaLeonova:


pkg/testutil/rand/model.go
```diff
@@ -125,7 +125,7 @@ func WithStatus(status model.MessageStatus) ModelMessageOption {
 // WithMessageFormat encloses the message in curly braces.
 func WithMessageFormat() ModelMessageOption {
```
klavinsjanis, line 126: is this function even used anywhere?

VladislavsPerkanuks (commented):
You had to change single curly braces to double, but you changed them to quadruple 🤔 
e.g.
```
{{ text }}
not
{{{{ text }}}}
```

PR #214, opened 2023-12-05
210-xliff2-to-mf2

klavinsjanis:


.golangci.yml
```diff
@@ -53,10 +53,19 @@ linters:
     - wastedassign
     - whitespace
     - zerologlint
+    - ireturn
```
VladislavsPerkanuks, line 56: why disable, and then configuring disabled linter ? 🤔

pkg/convert/xliff2.go
```diff
@@ -1,17 +1,38 @@
...
+	ph  = "ph"  // standalone code of the original format
+	pc  = "pc"  // well-formed spanning original code
+	sc  = "sc"  // start of a spanning original code
+	ec  = "ec"  // end of a spanning original code
+	mrk = "mrk" // annotation pertaining to the marked span
+	sm  = "sm"  // start marker of an annotation
+	em  = "em"  // end marker of an annotation
+)
 
 type xliff2 struct {
```
VladislavsPerkanuks, line 37: i think structs fields have to be unexported

pkg/convert/xliff2.go
```diff
@@ -20,23 +41,34 @@ type xliff2 struct {
...
 	Units []unit `xml:"unit"`
 }
 
 type unit struct {
-	ID     string  `xml:"id,attr"`                  // translation.messages[n].ID
-	Notes  *[]note `xml:"notes>note"`               // Set as pointer to avoid empty <notes></notes> when marshalling.
-	Source string  `xml:"segment>source"`           // translation.messages[n].Message (if no target language is set)
-	Target string  `xml:"segment>target,omitempty"` // translation.messages[n].Message (if target language is set)
-	// No unified standard about storing fuzzy values
+	ID    string  `xml:"id,attr"`    // TODO: translation.messages[n].ID should be based on source text.
```
VladislavsPerkanuks, line 50: why TODO ?
Message.ID == unit.ID

Message.ID is a identifier it can be source text, like in PO files, but in this case, XLIFF, has their own identifier

pkg/convert/xliff2.go
```diff
@@ -145,6 +140,111 @@ func ToXliff2(translation model.Translation) ([]byte, error) {
...
 
+// messageFromUnit converts Xliff2.0 unit element to model.Message.
+func messageFromUnit(u unit, original bool) (*model.Message, error) {
+	var (
+		err     error
+		decoder *xml.Decoder
+	)
+
+	m := &model.Message{
+		ID:          u.ID, // TODO: translation.messages[n].ID should be based on source text.
```
VladislavsPerkanuks, line 151: Same,redundant TODO

VladislavsPerkanuks: If you want to preserve original text, I can suggest adding local variable, containing that text.

e.g.
```go
if !original{
  message.Local($sourceText, mf2.Literal(...))
}
```

.golangci.yml
klavinsjanis: oh, you right, I thought it was enable :)

pkg/convert/xliff2.go
klavinsjanis: aren't they exported already?

klavinsjanis: Well, for Unmarshal to struct we need exported.

pkg/convert/xliff2.go
klavinsjanis: Do we have to keep original text though?

pkg/convert/xliff2.go
VladislavsPerkanuks: sure, forgot :)

pkg/convert/xliff2.go
VladislavsPerkanuks: we lose the source tag when we convert back nonoriginal translation -> xliff.
However, I am unsure if it is important for apps, which utilize Xliff...

pkg/convert/xliff2.go
```diff
@@ -145,6 +140,116 @@ func ToXliff2(translation model.Translation) ([]byte, error) {
...
+			case cp, ph, sc, ec, sm, em, pc, mrk:
+				elementCount[elName]++
+
+				v := "$" + elName + strconv.Itoa(elementCount[elName])
+				message.Expr(mf2.Var(v))
+
+				var attributes string
+
+				for i := range t.Attr {
+					attributes += fmt.Sprintf(` %s="%s"`, t.Attr[i].Name.Local, t.Attr[i].Value)
```
janishorsts, line 204: Why not marshalling using XML? Are you sure it always serializes correctly to XML?

janishorsts: Why do we store XML syntax in MF2?

pkg/convert/xliff2.go
```diff
@@ -145,6 +140,116 @@ func ToXliff2(translation model.Translation) ([]byte, error) {
 
 // helpers
 
+// messageFromUnit converts Xliff2.0 unit element to model.Message.
+func messageFromUnit(u unit, original bool) (*model.Message, error) {
```
janishorsts, line 144: what problem does this solve?

pkg/convert/xliff2.go
klavinsjanis: Why do we store XML syntax in MF2? 
Followed design document that @VladislavsPerkanuks  drew up. Maybe he can elaborate.

pkg/convert/xliff2.go
klavinsjanis: // messageFromUnit converts Xliff2.0 unit element to model.Message.

pkg/convert/xliff2.go
VladislavsPerkanuks: in MF2 we store everything, that is related to message and is important, and cannot fit into model.Message struct.
So we store them as locals. When converting back to xliff, we get that info from locals, and store them back into xliff struct.

That was the way, that @janishorsts and me agreed on.

pkg/convert/xliff2.go
janishorsts: Examples?

janishorsts:
Too many unknowns. Attempt to use constructs modern applications don't have. Rework in the future.

PR #217, opened 2023-12-13
178-make-repo-transactional

klavinsjanis:


pkg/repo/badgerdb/repo.go
```diff
@@ -11,6 +11,7 @@ import (
 // Repo implements the repo interface.
 type Repo struct {
 	db *badger.DB
+	tx *badger.Txn
```
janishorsts, line 14: I believe we can use a single field as the interface instead of a concrete type. Can you check what interfaces are defined in the `badger` and `sql`?

E.g., it does not matter whether we use `badger.Txn` or `badger.DB` in the `db` field.

pkg/repo/mysql/translation.go
```diff
@@ -13,23 +13,64 @@ import (
...
+
 func (r *Repo) SaveTranslation(ctx context.Context, serviceID uuid.UUID, translation *model.Translation) error {
-	_, err := r.LoadService(ctx, serviceID)
-	if err != nil {
-		return fmt.Errorf("repo: load service: %w", err)
+	if _, ok := r.db.(*sql.Tx); ok { // use existing tx
+		return r.saveTranslation(ctx, serviceID, translation)
 	}
 
-	tx, err := r.db.BeginTx(ctx, nil)
```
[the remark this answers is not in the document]
VladislavsPerkanuks: Correct me if I am wrong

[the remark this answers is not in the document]
VladislavsPerkanuks: i guess i am wrong :)
nvm then

VladislavsPerkanuks (approved):

pkg/repo/badgerdb/translation.go
```diff
@@ -20,27 +20,54 @@ func translationKey(serviceID uuid.UUID, language language.Tag) []byte {
...
+
+		return fmt.Errorf("repo: execute tx: %w", err)
+	}
+
+	if err := tx.Commit(); err != nil {
+		return fmt.Errorf("repo: commit tx: %w", err)
+	}
+
+	return nil
+}
```
janishorsts, line 44: When `fn()` panics, no error is returned - false positive. 

Let's return an error when it panics.

PS. The code is self-descriptive, there's no need for a comment that we start a tx. :)


```suggestion
func (r *Repo) Tx(ctx context.Context, fn func(repo.TranslationsRepo) error) (err error) {
	tx := r.db.NewTransaction(true)

	defer func() {
		if r := recover(); r != nil {
			tx.Discard()
			err = fmt.Errorf("repo: tx panicked: %v", r)
		}
	}()

	if err = fn(&Repo{db: r.db, tx: tx}); err != nil {
		tx.Discard()

		return fmt.Errorf("repo: execute tx: %w", err)
	}

	if err = tx.Commit(); err != nil {
		return fmt.Errorf("repo: commit tx: %w", err)
	}

	return nil
}
```

janishorsts: Let's pass ctx, too.

```go
func (r *Repo) Tx(ctx context.Context, fn func(ctx context.Context, repo.TranslationsRepo) error) (err error) {
```

pkg/repo/mysql/translation.go
```diff
@@ -13,23 +13,64 @@ import (
 	"golang.org/x/text/language"
 )
 
+// Tx executes a transaction on the repository.
+// Returns an error if there was an error starting the transaction,
+// executing the callback function, or committing the transaction.
+func (r *Repo) Tx(ctx context.Context, fn func(repo.TranslationsRepo) error) error {
```
janishorsts, line 19: The same here - return error on panic. And let's pass ctx too.

pkg/repo/badgerdb/translation.go
```diff
@@ -20,27 +20,54 @@ func translationKey(serviceID uuid.UUID, language language.Tag) []byte {
 	return []byte(fmt.Sprintf("%s%s:%s", translationPrefix, serviceID, language))
 }
 
+func (r *Repo) Tx(ctx context.Context, fn func(repo.TranslationsRepo) error) error {
```
janishorsts, line 23: Doesn't it belong to `repo.go`? `Tx` has nothing to do with the `Translation` resource.

pkg/repo/repo.go
```diff
@@ -25,6 +25,8 @@ type LoadTranslationsOpts struct {
 }
 
 type TranslationsRepo interface {
+	// Tx executes fn in a transaction.
+	Tx(ctx context.Context, fn func(TranslationsRepo) error) error
```
janishorsts, line 29:
```suggestion
	Tx(ctx context.Context, fn func(ctx context.Context, repo TranslationsRepo) error) error
```

pkg/server/file.go
```diff
@@ -153,18 +153,24 @@ func (t *TranslateServiceServer) UploadTranslationFile(
...
 	// Update affected translations
-	for i := range all {
-		err = t.repo.SaveTranslation(ctx, params.serviceID, &all[i])
-
-		switch {
-		case errors.Is(err, repo.ErrNotFound):
-			return nil, status.Errorf(codes.NotFound, "service not found")
-		case err != nil:
-			return nil, status.Errorf(codes.Internal, "")
+	txErr := t.repo.Tx(ctx, func(tr repo.TranslationsRepo) error {
```
janishorsts, line 156: Why not?

```suggestion
	err = t.repo.Tx(ctx, func(tr repo.TranslationsRepo) error {
```

pkg/repo/badgerdb/translation.go
klavinsjanis: Hmm, got confused by the TranslationsRepo naming. Why aren't we declaring all method signatures in the Repo interface directly?

janishorsts: We talk about different things. I guess.

The method `Tx()` belongs to the Repo implementation for each database - `pkg/repo/<db>/repo.go`. You put the implementation in the `pkg/repo/badgerdb/translation.go` file that MUST implement only methods for the Translation resource. Why did you put it in this file?

You talk about the interfaces in the `pkg/repo/repo.go` file where they are grouped by resource. `Tx` MUST be declared in the `Repo` interface.

pkg/repo/badgerdb/translation.go
```diff
@@ -22,25 +22,28 @@ func translationKey(serviceID uuid.UUID, language language.Tag) []byte {
 
 // SaveTranslation handles both Create and Update.
 func (r *Repo) SaveTranslation(ctx context.Context, serviceID uuid.UUID, translation *model.Translation) error {
+	if r.tx != nil { // use existing tx
+		return r.saveTranslation(ctx, serviceID, translation)
+	}
+
+	return r.Tx(ctx, func(ctx context.Context, rp repo.Repo) error { // create new tx
+		return rp.SaveTranslation(ctx, serviceID, translation) //nolint:wrapcheck
+	})
```
janishorsts, line 31: Create a private and reusable helper function to ensure we have a tx. It helps to reduce boilerplate code when implementing Repo methods that require tx.

```suggestion
	return r.ensureTx(ctx, func(ctx context.Context, rp repo.Repo) error { // create new tx
		return rp.SaveTranslation(ctx, serviceID, translation) //nolint:wrapcheck
	})
```

janishorsts: Please note, that we don't need `saveTranslation` anymore.

pkg/repo/mysql/translation.go
```diff
@@ -14,22 +14,25 @@ import (
...
+		return r.saveTranslation(ctx, serviceID, translation)
 	}
 
-	tx, err := r.db.BeginTx(ctx, nil)
+	return r.Tx(ctx, func(ctx context.Context, rp repo.Repo) error { // create new tx
+		return rp.SaveTranslation(ctx, serviceID, translation) //nolint:wrapcheck
+	})
+}
+
+func (r *Repo) saveTranslation(ctx context.Context, serviceID uuid.UUID, translation *model.Translation) error {
```
janishorsts, line 26: The same here, `saveTranslation` is not required, called only once in `SaveTranslation`.

janishorsts (changes requested):

pkg/repo/mysql/translation.go
```diff
@@ -14,94 +14,85 @@ import (
...
+		// update message's message, description and status values.
+		stmt, err := r.db.PrepareContext(
 			ctx,
-			translationID,
-			m.ID,
-			m.Message,
-			m.Description,
-			m.Positions,
-			m.Status,
+			`INSERT INTO message
```
janishorsts, line 64: Please ensure the target consumer of this SQL statement is an observability user. This applies to ALL projects we work on.

Not a developer.

janishorsts (approved):

PR #219, opened 2023-12-28
update translation message

DarjaLeonova:


cmd/translate/main_grpc_integration_test.go
```diff
@@ -684,33 +686,43 @@ func Test_UpdateTranslation_gRPC(t *testing.T) {
...
 		expectedCode codes.Code
 	}{
 		{
-			name:         "Happy Path",
+			name:         "Happy Path update all",
 			request:      happyReq,
 			expectedCode: codes.OK,
 		},
+		{
+			name:         "Happy path update original field",
```
klavinsjanis, line 722: update_mask 'original'? What about update_mask targeting 'messages' field?

pkg/model/mask.go
```diff
@@ -135,3 +157,17 @@ func UpdateService(src, dst *Service, mask Mask) error {
...
 }
+
+// UpdateTranslation updates the destination translation based on the source translation and field mask.
+func UpdateTranslation(src, dst *Translation, mask Mask) error {
+	// Prevent updating read-only fields like ID
+	if slices.ContainsFunc(mask, func(s string) bool {
+		return strings.EqualFold(s, "ID")
+	}) {
+		return errors.New("\"id\" is not allowed in field mask")
+	}
```
VladislavsPerkanuks, line 168: Translation does not have any read-only fields, and also it does not have ID.
so possible values for mask is
```go
language
messages
original
```

pkg/model/mask.go
```diff
@@ -115,6 +110,33 @@ func updateField(src, dst reflect.Value, fields []string) {
 	}
 }
 
+// updateSliceField this function updates a destination slice by either replacing a struct with a matching ID
+// or appending new values if no matching ID is found.
+func updateSliceField(srcField, dstField reflect.Value) {
```
klavinsjanis, line 115: Maybe we should consider using a custom implementation for update_mask 'messages' since it might be used frequently.

pkg/model/mask.go
```diff
@@ -135,3 +157,17 @@ func UpdateService(src, dst *Service, mask Mask) error {
 
 	return nil
 }
+
+// UpdateTranslation updates the destination translation based on the source translation and field mask.
+func UpdateTranslation(src, dst *Translation, mask Mask) error {
```
VladislavsPerkanuks, line 162: I would go with following logic, that would simplify everything

1. Check mask, if it does not contain messages, then just call update()
2. If it exists, then implement logic of updating slice here, not in update() function. Pop the messages from mask, and call update, if mask is not empty yet

cmd/translate/main_grpc_integration_test.go
VladislavsPerkanuks: does not matter, as this test do not check, if implementation doing the right thing, it just checks that request gone through and returned ok code.

tho It could test, not only the status code, but also the returned value

cmd/translate/main_rest_integration_test.go
```diff
@@ -640,22 +640,22 @@ func Test_UpdateTranslation_REST(t *testing.T) {
 	createTranslation(ctx, t, service.GetId(), &translatev1.Translation{Original: true, Language: langs[0].String()})
 	createTranslation(ctx, t, service.GetId(), &translatev1.Translation{Original: false, Language: langs[1].String()})
 
-	happyReq := randUpdateTranslationReq(t, service.GetId(), &translatev1.Translation{Language: langs[1].String()})
+	happyReq := randUpdateTranslationReq(t, service.GetId(), &translatev1.Translation{Language: langs[1].String()}, nil)
```
VladislavsPerkanuks, line 643: missing test case with field mask for REST

pkg/model/mask_test.go
```diff
@@ -299,3 +301,92 @@ func Test_UpdateServiceFromMask(t *testing.T) {
...
+				require.Equal(t, original.Language, dstTranslation.Language)
+				require.Equal(t, original.Original, dstTranslation.Original)
+			},
+		},
+		{
+			name:        "Try to update ID",
+			fieldMask:   Mask{"ID"},
+			expectedErr: errors.New("\"id\" is not allowed in field mask"),
+		},
+	}
```
VladislavsPerkanuks, line 372: there is no id in tranlsation resource

pkg/server/translation.go
```diff
@@ -248,6 +253,12 @@ func (t *TranslateServiceServer) UpdateTranslation(
 		}
 	}
 
+	for i := range all {
+		if err = model.UpdateTranslation(params.translation, &all[i], params.mask); err != nil {
+			return nil, status.Errorf(codes.InvalidArgument, err.Error())
+		}
+	}
```
VladislavsPerkanuks, line 260: That will break things.

Scenario
```go
// in DB
// Translation1 with Language LV
// Translation2 with Language En

// Request to update Translation1 language with new language RU

// The result of this selected code should be
// Translation1 with Language RU
// Translation2 with Language RU

// And then saving to repo, will try to save multiple translation into one composite key (service + language)
```

pkg/model/mask.go
janishorsts: When we have a slice with a struct element with `ID` field, we assume it is the unique identifier and we can update this element.

What update mask do we use to update the Message in Translation? `messages.*`?

janishorsts: https://github.com/protocolbuffers/protobuf/blob/9bbea4aa65bdaf5fc6c2583e045c07ff37ffb0e7/src/google/protobuf/field_mask.proto#L111

```proto
// If a repeated field is specified for an update operation, new values will
// be appended to the existing repeated field in the target resource. Note that
// a repeated field is only allowed in the last position of a `paths` string.
```

I assume, that when the update mask is a repeated field, we append by default. However, if the repeated message contains the field `id` (unique identifier), we replace it, right?

In our case, update mask `messages` replaces the message by its id:
* appends if it is a new message
* replaces if it is an existing message

pkg/model/mask.go
janishorsts: We should ignore managing read-only fields for now and determine general-purpose and reusable solutions later.

It should be closer to the request validation. Events (gRPC etc) that have reached inside our system MUST be already safe to process.

```go
// allowMask returns true if updateMask does not contain any readOnlyMask.
func allowMask(updateMask, readOnlyMask Mask) bool
```

pkg/model/mask.go
DarjaLeonova: @janishorsts Currently the updatesSliceField function appends a new message if ID doesn't exist and replaces a message if ID exists. You can check tests in mask_test.go. So as I understood it is expected behavior, right?

pkg/model/mask.go
VladislavsPerkanuks: there is no mention of read-only fields in my comment

janishorsts: We don't need to define `func Update<resource>` for each resource. We can get away with general-purpose functions.

pkg/model/mask.go
```diff
@@ -135,3 +160,43 @@ func UpdateService(src, dst *Service, mask Mask) error {
...
+	// nonMsgField will store the field paths that are not related to "Messages."
+	nonMsgField := []string{}
+
+	for _, path := range mask {
+		switch path {
+		case "Messages":
```
janishorsts, line 171: This works only for Translation.

How can you implement this as a general purpose solution in `updateField`? Based on https://github.com/translate-agent/translate/pull/219/files#r1437547444

`when the update mask is a repeated field, we append by default. However, if the repeated message contains the field id (unique identifier), we replace it`

pkg/model/mask.go
```diff
@@ -115,6 +110,50 @@ func updateField(src, dst reflect.Value, fields []string) {
...
+		_, ok := el.FieldByName("ID")
+
+		return !ok
+	}
+
+	if appendOnly() {
+		// If the field is a slice, append new values from src to existing slice in dst
+		//nolint:lll
+		// https://github.com/protocolbuffers/protobuf/blob/9bbea4aa65bdaf5fc6c2583e045c07ff37ffb0e7/src/google/protobuf/field_mask.proto#L111
+		dstField.Set(reflect.AppendSlice(dstField, srcField))
```
janishorsts, line 132: We can remove this comment but move the link to protobuf description to func description.
```suggestion
		dstField.Set(reflect.AppendSlice(dstField, srcField))
```

pkg/model/mask.go
```diff
@@ -115,6 +110,50 @@ func updateField(src, dst reflect.Value, fields []string) {
...
+				found = true
+			}
+		}
+
+		if !found {
+			dstField.Set(reflect.Append(dstField, srcField.Index(i)))
+			continue
+		}
+	}
+}
```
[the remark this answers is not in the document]
VladislavsPerkanuks: i would go with
```go

func findInDstField(srcID interface{}, dstField reflect.Value) (int, bool) {
	for i := 0; i < dstField.Len(); i++ {
			if dstField.Index(j).FieldByName("ID").Interface() == srcID {
					return i
			}
	}
	return -1
}

...

srcID := srcField.Index(i).FieldByName("ID").Interface()
idx := findInDstField(srcID, dstField)

// idx == -1 -> append
// idx != -1 -> replace

```

pkg/model/mask.go
```diff
@@ -115,6 +110,50 @@ func updateField(src, dst reflect.Value, fields []string) {
 	}
 }
 
+// updateSliceField this function updates a destination slice by either replacing a struct with a matching ID
+// or appending new values if no matching ID is found.
+func updateSliceField(srcField, dstField reflect.Value) {
+	appendOnly := func() bool {
```
VladislavsPerkanuks, line 116: A litle more simple
```suggestion
appendOnly := func() bool {
    el := srcField.Type().Elem()
    _, ok := el.FieldByName("ID")
     
     return el.Kind() != reflect.Struct || !ok
    }
```

pkg/model/mask.go
```diff
@@ -115,6 +110,50 @@ func updateField(src, dst reflect.Value, fields []string) {
...
+		el := srcField.Type().Elem()
+
+		if el.Kind() != reflect.Struct {
+			return true
+		}
+
+		_, ok := el.FieldByName("ID")
+
+		return !ok
```
VladislavsPerkanuks, line 125: A litle more simple
```suggestion
appendOnly := func() bool {
    el := srcField.Type().Elem()
    _, ok := el.FieldByName("ID")
     
     return el.Kind() != reflect.Struct || !ok
    }
```

pkg/model/mask.go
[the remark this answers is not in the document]
janishorsts: https://pkg.go.dev/slices#IndexFunc

pkg/model/mask.go
janishorsts: We don't need `UpdateTranslation` anymore.

Let's expose `update` and use it instead.

proto/translate/v1/translate.proto
```diff
@@ -83,6 +83,7 @@ message UpdateTranslationRequest {
   string service_id = 1;
   bool populate_translations = 2;
   Translation translation = 3;
+  google.protobuf.FieldMask update_mask = 4;
 }
```
janishorsts, line 87: Let's change the order of fields in the proto message to make it more logical order.

PS. This cannot be done when the code is in the production.

```suggestion
  string service_id = 1;
  Translation translation = 2;
  google.protobuf.FieldMask update_mask = 3;
  bool populate_translations = 4;
}
```

cmd/translate/main_rest_integration_test.go
```diff
@@ -640,22 +641,32 @@ func Test_UpdateTranslation_REST(t *testing.T) {
...
+
+	req := randUpdateTranslationReq(t, service.GetId(), &translatev1.Translation{
+		Language: langs[1].String(),
+		Messages: []*translatev1.Message{
+			{
+				Id: "Hello", Message: "World",
+			},
+		},
+	},
+		&field_mask.FieldMask{Paths: []string{"messages"}})
```
klavinsjanis, line 654: Maybe could add same test to grpc client integration tests as well.

pkg/model/mask.go
DarjaLeonova: This solution will cause a panic. We should ensure that the type on which we are calling FieldByName is indeed a struct type.

VladislavsPerkanuks: oki

pkg/server/translation.go
```diff
@@ -248,6 +253,13 @@ func (t *TranslateServiceServer) UpdateTranslation(
 		}
 	}
 
+	if err = model.UpdateTranslation(
```
klavinsjanis, line 256: Could you add an integration test where a message for an original translation is updated/appended ? 
So that we know everything is working as intended.

VladislavsPerkanuks (approved):
From my perspective looks fine. Fix other people's comments and should be good.

And let's hope it does not bring any unintended behavior into the system 🙈

pkg/model/mask.go
```diff
@@ -115,6 +110,53 @@ func updateField(src, dst reflect.Value, fields []string) {
...
+			// If found, update the corresponding structure in the destination slice
+			dstField.Index(idx).Set(srcField.Index(i))
+			continue
+		}
+
+		dstField.Set(reflect.Append(dstField, srcField.Index(i)))
+	}
+}
+
+func findInDstField(srcID interface{}, dstField reflect.Value) int {
```
janishorsts, line 150: `func updateSliceField` is relatively simple.

Can we add this the same way you did with `appendOnly` to avoid package level functions?

Simply `index := func(id any) int {}`?

pkg/model/mask.go
janishorsts: It is less performant, too.

janishorsts: @VladislavsPerkanuks I like the thinking, though. But check my previous comment. :)

proto/translate/v1/translate.proto
```diff
@@ -150,6 +151,8 @@ service TranslateService {
     };
   }
 
+  //nolint:lll
```
janishorsts, line 154: `.proto` is not golang source code, isn't it? `//nolint:` is applicable only the `.go` files.

proto/translate/v1/translate.proto
```diff
@@ -150,6 +151,8 @@ service TranslateService {
...
+  // https://github.com/protocolbuffers/protobuf/blob/9bbea4aa65bdaf5fc6c2583e045c07ff37ffb0e7/src/google/protobuf/field_mask.proto#L111
```
janishorsts, line 155: Why did you add this comment here?

pkg/model/mask.go
```diff
@@ -115,6 +110,53 @@ func updateField(src, dst reflect.Value, fields []string) {
...
+	// If found by id, replace the corresponding structure in the destination slice. Otherwise, append.
+	for i := 0; i < srcField.Len(); i++ {
+		srcID := srcField.Index(i).FieldByName("ID").Interface()
+		idx := findInDstField(srcID, dstField)
+
+		// Check if the "ID" field of the first element in the source slice
+		// matches the "ID" field of the current element in the destination slice.
+		if idx != -1 {
+			// If found, update the corresponding structure in the destination slice
+			dstField.Index(idx).Set(srcField.Index(i))
```
janishorsts, line 142: The code is self-descriptive. No need for comments.

My personal preference, emphasizes all valid values, inclusive range.

```suggestion
		if idx >= 0 {
			dstField.Index(idx).Set(srcField.Index(i))
```
Emphasizes all valid values, exclusive range.

```suggestion
		if idx > -1 {
			dstField.Index(idx).Set(srcField.Index(i))
```

klavinsjanis (approved):

klavinsjanis:
Did you merge in origin/main?

cmd/translate/main_grpc_integration_test.go
```diff
@@ -82,7 +84,7 @@ func createService(ctx context.Context, t *testing.T) *translatev1.Service {
 // createTranslation creates a random translation, and calls the CreateTranslation RPC.
 func createTranslation(ctx context.Context, t *testing.T, serviceID string,
 	override *translatev1.Translation,
-) *translatev1.Translation { //nolint:unparam
+) *translatev1.Translation { //nolintlint:unparam
```
klavinsjanis, line 87: is this nolint needed?

DarjaLeonova: Looks like //nolint:unparam is not needed at all ;D

cmd/translate/main_grpc_integration_test.go
```diff
@@ -672,6 +674,272 @@ func Test_ListTranslations_gRPC(t *testing.T) {
...
+	service7 := createService(ctx, t)
+	service8 := createService(ctx, t)
+
+	tests := []struct {
+		request     *translatev1.UpdateTranslationRequest
+		translation *translatev1.Translation
+		name        string
+	}{
+		{
+			name: "Update messages field in original translation",
```
klavinsjanis, line 716: Could you group these tests? 
Right now they are a. bit mixed up.
Something like -
update message/s for translation, 
update message/s for original, 
update original status for translation

cmd/translate/main_grpc_integration_test.go
```diff
@@ -672,6 +674,272 @@ func Test_ListTranslations_gRPC(t *testing.T) {
...
+		tt := tt
+		subtest(tt.name, func(ctx context.Context, t *testing.T) {
+			beforeUpdateTranslations, err := client.ListTranslations(ctx, &translatev1.ListTranslationsRequest{
+				ServiceId: tt.request.GetServiceId(),
+			})
+
+			require.NoError(t, err)
+			resp, err := client.UpdateTranslation(ctx, tt.request)
+
+			if err == nil {
```
klavinsjanis, line 914: why do you check if err is nil? Isn't all of these happy path tests?

cmd/translate/main_grpc_integration_test.go
```diff
@@ -672,6 +674,272 @@ func Test_ListTranslations_gRPC(t *testing.T) {
...
+			afterUpdateTranslations, err := client.ListTranslations(ctx, &translatev1.ListTranslationsRequest{
+				ServiceId: tt.request.GetServiceId(),
+			})
+
+			require.NoError(t, err)
+			require.NotEmpty(t, resp)
+
+			var translationFromService *translatev1.Translation
+
+			for i := range afterUpdateTranslations.GetTranslations() {
```
janishorsts, line 927: Never let actual values guide your tests. Use your expectations to guide.

E.g. `ListTranslations` returns an empty list, the test passes, right? But, actually, it should fail.

cmd/translate/main_grpc_integration_test.go
```diff
@@ -672,6 +674,272 @@ func Test_ListTranslations_gRPC(t *testing.T) {
...
+			request: randUpdateTranslationReq(t, service4.GetId(), &translatev1.Translation{
+				Language: langs[0].String(),
+				Messages: []*translatev1.Message{
+					{
+						Id: "Hello", Message: "World",
+					},
+				},
+			},
+				&field_mask.FieldMask{Paths: []string{"messages"}}),
+			translation: createTranslation(ctx, t, service4.GetId(), &translatev1.Translation{
```
janishorsts, line 789: All tests MUST be isolated - no resource sharing. Any reason why you use the same service as in the previous test?

DarjaLeonova: I will fix it. 👍

pkg/model/mask.go
```diff
@@ -54,7 +54,7 @@ type Mask []string
 //
 //	update(&src, &dst, mask)
```
mvilks, line 55: Uppercase

mvilks (approved):

pkg/model/mask.go
```diff
@@ -115,6 +110,51 @@ func updateField(src, dst reflect.Value, fields []string) {
...
+		}
+
+		index := idx(srcID)
+
+		if index >= 0 {
+			dstField.Index(index).Set(srcField.Index(i))
+			continue
+		}
+
+		dstField.Set(reflect.Append(dstField, srcField.Index(i)))
```
VladislavsPerkanuks, line 154:
```suggestion

		switch j := idx(srcID); j {
		default: // found -> replace
			dstField.Index(j).Set(srcField.Index(i))
		case -1: // not found -> append
			dstField.Set(reflect.Append(dstField, srcField.Index(i)))
		}
```

VladislavsPerkanuks (approved):

pkg/model/mask.go
janishorsts:
```suggestion
//	Update(&src, &dst, mask)
```

pkg/model/mask.go
```diff
@@ -33,7 +33,7 @@ import (
 //	...
 type Mask []string
 
-// update updates the dst with the values from src based on the mask.
+// Update updates the dst with the values from src based on the mask.
```
janishorsts, line 36: Please describe how slices are handled in the exposed func.

```go
//
// When updating slices:
//   - replace the element if slice contains a struct with a field "ID"
//   - append in all other cases
```

PR #227, opened 2024-01-11
Convert Translation to PO

VladislavsPerkanuks:
After this PR we should have a working implementation for PO files, PO -> Translation -> PO

**What is done:**
- Rewritten `Translation->PO` implementation
- Added `PO.Marshal(),` to convert PO struct to `[]byte`
- Deleted all old tests` (PO -> Translation)`
- Merged Convert Tests into one.` PO -> Translation -> PO`. That way we are 100% sure that converting back and forth does not change anything in between. Also, `FromPo` and `ToPo` work as intended. 

**What is not done**
- Plural forms are not preserved l in the transformation, but they are crucial to apps that use PO. We need to think of a way to preserve that.
- Renaming all `pot` references to `po`, packages, references, files, ..., as it is more correct

**Note**
~Due to renaming + making changes in the file, git does not think it is the same file (old: pot.go, pot_test.go new: po.go po_test.go), hence the diff is quite big~
**Revert to original pkg and file names (pot) as git, does not like it. (git diff instead renamed+changes, says new file added, old file deleted, which increased PR diff size dramasticly)**


pkg/pot/parse.go
```diff
@@ -35,11 +65,75 @@ type MessageNode struct {
...
+	return sb.Bytes()
+}
+
 type Po struct {
 	Header   HeaderNode
 	Messages []MessageNode
 }
 
+// Marshal serializes the Po object into a byte slice.
+func (p *Po) Marshal() []byte {
```
klavinsjanis, line 125: Shouldn't Matshal return an error as well? Idiomatic for golang.

VladislavsPerkanuks: nothing to error there

pkg/server/convert.go
```diff
@@ -62,7 +62,7 @@ func TranslationToData(schema translatev1.Schema, translation *model.Translation
 	case translatev1.Schema_JSON_NGX_TRANSLATE:
 		to = convert.ToNgxTranslate
 	case translatev1.Schema_POT:
```
janishorsts, line 64:
```suggestion
	case translatev1.Schema_PO:
```

pkg/convert/pot.go
```diff
@@ -140,11 +133,18 @@ func msgNodeToMF2(node pot.MessageNode, getMessages func(pot.MessageNode) []stri
 		return nil
 	}
 
+	// look for format flag, e.g. #, python-format, c-format, no-python-format, etc.
+	formatFlagIdx := slices.IndexFunc(node.Flags, func(flag string) bool {
+		return strings.Contains(flag, "-format")
```
janishorsts, line 138: I don't know the flag's value, but shouldn't we check the suffix?

```suggestion
		return strings.HasSuffix(flag, "-format")
```

pkg/server/convert.go
VladislavsPerkanuks: I reverted back, all changes regarding renaming pot -> po. in pkgs, references, files, etc.
git diff was acting silly.

better to do that in separate PR

pkg/convert/pot.go
VladislavsPerkanuks: you are right

pkg/convert/pot.go
```diff
@@ -236,271 +234,98 @@ func textWithPlaceholders(mfBuilder *mf2.Builder, msg string, placeholders map[s
...
-	if err := write(b, tag, str); err != nil {
-		return fmt.Errorf("write: %w", err)
+	// If original, msgstr are empty
+	if t.Original {
+		patternsToMsg = func(patterns []ast.Pattern) string { return "" }
 	}
 
-	return nil
-}
+	unquoteLiteral := func(l ast.Literal) string { return strings.ReplaceAll(l.String(), "|", "") }
```
janishorsts, line 268: We'll need to resolve how we handle raw and MS2 formatted values. Not part of this PR.

pkg/pot/parse.go
```diff
@@ -16,11 +17,33 @@ type HeaderNode struct {
 	PluralForms PluralForm
 }
 
+func (h *HeaderNode) marshal(b *bytes.Buffer) {
+	b.WriteString("msgid \"\"\nmsgstr \"\"\n")
+
+	if h.Language != language.Und {
+		b.WriteString(fmt.Sprintf("\"Language: %s\\n\"\n", h.Language.String()))
```
janishorsts, line 24: alternative

```suggestion
		b.WriteString(fmt.Sprintf("\"Language: %v\\n\"\n", h.Language))
```

janishorsts (approved):

pkg/pot/parse.go
VladislavsPerkanuks: yeah, why not

pkg/convert/pot.go
VladislavsPerkanuks: created issue so we wont forget

pkg/server/convert.go
VladislavsPerkanuks: created issue #229

PR #230, opened 2024-01-12
226-google-translate-mf2

klavinsjanis:


pkg/fuzzy/fuzzy_integration_test.go
```diff
@@ -20,14 +20,12 @@ import (
 // ---–––--------------Actual Tests------------------–––---
 
 func Test_Translate(t *testing.T) {
-	// NOTE: Tests skipped for now until fuzzy translation is fixed.
-	t.Skip()
 	t.Parallel()
 
 	allTranslators(t, func(t *testing.T, translator Translator, subTest testutil.SubtestFn) {
```
VladislavsPerkanuks, line 25: Few remarks about fuzzy tests:
1. this test does not check if returned value is still correct mf2 message.  For that I would suggest use mock testing, where translate client returns same translatable input. That way you could check Translate() logic.
2. in integration test one case would be enough, either it goes through or it doesn't. Translate method's logic can be tested using mocks
3. random data is not required, as we want that test specific cases

pkg/fuzzy/google_translate.go
```diff
@@ -113,83 +117,46 @@ func (g *GoogleTranslate) Translate(
...
+			return nil, fmt.Errorf("google translate client: translate text #%d from batch: %w", i, err)
+		}
+
+		for i := range res.GetTranslations() {
+			translatedTexts = append(translatedTexts, res.GetTranslations()[i].GetTranslatedText())
+		}
+	}
+
+	// build translation with new translated text
+	translated, err := buildTranslated(translation, translatedTexts, targetLanguage)
```
VladislavsPerkanuks, line 156: I think this could be simplified by, not working with AST, but working with string.

we have original mf2 and translated text from mf2 with placeholdres $0. 

if is simple message, then message[i] = translatedText[i]

for complex
1. We capture original mf2 expression { $var } or something like that, lets call it `var ph1` string
2. We find index on where the actual message starts, lets call it idx
3. message = translatedTexts[i], with all $0 $1 replaced with ph1, ph2

pkg/server/file.go
```diff
@@ -100,6 +100,8 @@ func (t *TranslateServiceServer) UploadTranslationFile(
 		return nil, status.Errorf(codes.InvalidArgument, err.Error())
 	}
 
+	// TODO: fail validation if message contains '{\$(0|[1-9]\d*)\}'.
```
VladislavsPerkanuks: there is no message, in upload params

pkg/fuzzy/google_translate.go
```diff
@@ -113,83 +117,45 @@ func (g *GoogleTranslate) Translate(
 		return nil, nil //nolint:nilnil
 	}
 
-	// TODO: Implement translation of message texts.
-
 	// NOTE: temporary fix to avoid failing tests.
```
VladislavsPerkanuks, line 120: i guess this is redundnat now ?

pkg/fuzzy/google_translate.go
```diff
@@ -204,9 +170,8 @@ func parent() string {
 	return fmt.Sprintf("projects/%s/locations/%s", projectID, location)
 }
 
-// TODO: Remove if not used.
 // textToBatches splits text into batches with predefined maximum amount of elements.
-func textToBatches(text []string, batchLimit int) [][]string { //nolint:unused
+func textToBatches(text []string, batchLimit int) [][]string {
```
VladislavsPerkanuks, line 174: batchlimit is const
```suggestion
func textToBatches(text []string) [][]string {
```

pkg/fuzzy/google_translate.go
```diff
@@ -230,3 +195,250 @@ func textToBatches(text []string, batchLimit int) [][]string { //nolint:unused
...
+//		TextPattern("Hello"),
+//		PlaceholderPattern{ Expression: LiteralExpression{Literal: QuotedLiteral("name")}},
+//		TextPattern(" "),
+//		PlaceholderPattern{ Expression: LiteralExpression{Literal: QuotedLiteral("lastName")}},
+//		TextPattern("!"),
+//	}
+//
+// Output:
+// "Hello {$0} {$1}!".
+func printPattern(pattern []ast.Pattern) string {
```
VladislavsPerkanuks, line 266: **ambiguos**
The term “print” is commonly used in to refer to outputting information to the console. This function doesn’t print anything; it returns a string.

klavinsjanis: We have func with similar naming in the go-mf2 pkg.

VladislavsPerkanuks:
two remarks:
1) after calling Translate() original should be false, for the resulting translation, in any case
2) All messages are getting translated, even those with status Translated, which should be not the case

pkg/fuzzy/google_translate.go
VladislavsPerkanuks: dont think this is good :)

pkg/fuzzy/google_translate.go
klavinsjanis: I think we can keep it, just the comment has to go.

pkg/fuzzy/fuzzy_test.go
```diff
@@ -19,8 +20,7 @@ import (
 // ---–––--------------Actual Tests------------------–––---
 
 func Test_TranslateMock(t *testing.T) {
```
VladislavsPerkanuks, line 22: Instead of random messages, I suggest concrete messages, with all mf2 types, Simple,Complex, Matcher.
Here are my tests, which should cover all (positive) cases.
With the Translate() functionality behaving as I expect it to behave
```go
func Test_TranslateMock(t *testing.T) {
	// NOTE: AWS skipped for now until fuzzy translation is fixed.
	t.Parallel()

	ctx := context.Background()

	allMocks(t, func(t *testing.T, mock Translator) {
		tests := []struct {
			translation *model.Translation
			targetLang  language.Tag
			name        string
		}{
			{
				name:       "Simple message",
				targetLang: language.Latvian,
				translation: &model.Translation{
					Language: language.English,
					Original: true,
					Messages: []model.Message{
						{
							ID:      "1",
							Message: "Hello World!",
							Status:  model.MessageStatusTranslated,
						},
					},
				},
			},
			{
				name:       "Complex message quoted pattern",
				targetLang: language.Latvian,
				translation: &model.Translation{
					Language: language.English,
					Original: false,
					Messages: []model.Message{
						{
							ID: "1",
							Message: `.local $format = { python-format }
.local $name = { |{name}| }
{{Hello, { $name }!}}`,
							Status: model.MessageStatusUntranslated,
						},
					},
				},
			},
			{
				name:       "Complex message matcher",
				targetLang: language.Latvian,
				translation: &model.Translation{
					Language: language.Russian,
					Original: true,
					Messages: []model.Message{
						{
							ID: "1",
							Message: `.local $format = { python-format }
.local $suggestion = { |%(suggestion)s| }
.local $undefinedParameter = { |%(undefinedParameter)s| }
.local $firstSuggestions = { |%(firstSuggestions)s| }
.local $lastSuggestion = { |%(lastSuggestion)s| }
.match { $count }
1 {{{ $suggestion } вместо \\"{ $undefinedParameter }\\"?}}
2 {{
{ $firstSuggestions } или { $lastSuggestion } вместо 
\\"{ $undefinedParameter }\\"?}}
* {{
{ $firstSuggestions } или { $lastSuggestion } вместо 
\\"{ $undefinedParameter }\\"?}}`,
							Status: model.MessageStatusTranslated,
						},
					},
				},
			},
			// ID1 remains as is, ID2 is sent to translation.
			{
				name:       "One Translated message one not",
				targetLang: language.Latvian,
				translation: &model.Translation{
					Language: language.English,
					Original: false,
					Messages: []model.Message{
						{
							ID:      "1",
							Message: "Hello World!",
							Status:  model.MessageStatusTranslated,
						},
						{
							ID:      "2",
							Message: "Hello World!",
							Status:  model.MessageStatusUntranslated,
						},
					},
				},
			},
		}

		for _, tt := range tests {
			tt := tt
			t.Run(tt.name, func(t *testing.T) {
				t.Parallel()

				translated, err := mock.Translate(ctx, tt.translation, language.Latvian)
				require.NoError(t, err)

				// Check the that the translated translation have the correct language.
				require.Equal(t, tt.targetLang, translated.Language, "want: %s, got: %s", tt.translation.Language, translated.Language) //nolint:lll

				// Check that it is not original
				require.False(t, translated.Original, "translation cannot be original")

				// Check that length matches.
				require.Len(t, translated.Messages, len(tt.translation.Messages), "message count mismatch")

				for i, actualMsg := range translated.Messages {
					require.NotEmpty(t, actualMsg.Message)
					// In mock, we return the same message, so it should be the same.
					require.Equal(t, tt.translation.Messages[i].Message, actualMsg.Message)

					expectedStatus := tt.translation.Messages[i].Status // default: fuzzy or translated
					// If the message is not original and it is not translated, then it should be fuzzy.
					if !tt.translation.Original && tt.translation.Messages[i].Status != model.MessageStatusTranslated {
						expectedStatus = model.MessageStatusFuzzy
					}

					require.Equal(t, expectedStatus, actualMsg.Status, "want: %s, got: %s", expectedStatus, actualMsg.Status) //nolint:lll
				}
			})
		}
	})
}
```

klavinsjanis: random already creates 2 types - simple, complex and complex with matcher.

VladislavsPerkanuks (approved):
if it works then it works

PR #231, opened 2024-01-16
PO conversion tidy

VladislavsPerkanuks:
- Change placeholder extraction logic, in some cases it could be that format flag is present, but no placeholders are present
- Fixed issue with lexer, when text had escaped quote at then end, and lexer trimmed it.


pkg/pot/lex.go
```diff
@@ -222,7 +222,7 @@ func parseMultilineValue(line string, tokens []Token) error {
 		TokenTypeHeaderLanguageTeam, TokenTypeHeaderLastTranslator, TokenTypeHeaderXGenerator,
 		TokenTypeHeaderReportMsgidBugsTo, TokenTypeHeaderMIMEVersion, TokenTypeHeaderContentType,
 		TokenTypeHeaderContentTransferEncoding, TokenTypeHeaderGeneratedBy:
-		lastToken.Value += "\n" + strings.Trim(line, `"`)
+		lastToken.Value += "\n" + strings.TrimPrefix(strings.TrimSuffix(line, "\""), "\"")
```
VladislavsPerkanuks, line 225: not the same as `strings.Trim`

DarjaLeonova (approved):

klavinsjanis (approved):

PR #235, opened 2024-01-18
234-xliff-to-mf2

DarjaLeonova:
https://docs.oasis-open.org/xliff/v1.2/os/xliff-core.html#Struct_InLine
Should we also support tags that weren't specified in @VladislavsPerkanuks design document? Asking because most of them are deprecated in XLIFF1.1, but can still be used in 1.2

For Now added only test cases, to get feedback that tests are correct. after confirmation will start work on the implementation.

### DIFF
--- a/pkg/convert/xliff12_test.go
+++ b/pkg/convert/xliff12_test.go
@@ -151,7 +151,7 @@ func Test_FromXliff12(t *testing.T) {
 						ID: "9204248378636247318",
 						Message: ".local $x1 = { |<x id=\"PH\" equiv-text=\"status.filename\"/>| }\n" +
 							".local .local $x2 = { |<x id=\"PH2\" equiv-text=\"status.filename2\"/>| }\n " +
-							"{{Document { $x1 } was added to paperless.}}",
+							"{{Document { $x1 } was added to { $x2 } paperless.}}",
 						Status:    model.MessageStatusTranslated,
 						Positions: []string{"src/app/app.component.ts:51"},
 					},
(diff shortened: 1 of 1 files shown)


pkg/convert/xliff12_test.go
```diff
@@ -74,61 +74,214 @@ func randXliff12(translation *model.Translation) []byte {
...
+			</file>	
+			</xliff>`),
+			expected: &model.Translation{
+				Original: true,
+				Language: language.English,
+				Messages: []model.Message{
+					{
+						ID: "9204248378636247318",
+						Message: ".local $x1 = { |<x id=\"PH\" equiv-text=\"status.filename\"/>| }\n" +
+							"{{Document { $x1 } was added to paperless.}}",
```
janishorsts, line 121: Should it be a named placeholder? Does MF2 support variable name with `.` - `status.filename`? `{{Document { $status.filename } was added to paperless.}}`

pkg/convert/xliff12_test.go
```diff
@@ -74,61 +74,214 @@ func randXliff12(translation *model.Translation) []byte {
...
+				xmlns="urn:oasis:names:tc:xliff:document:1.2">
+			<file source-language="en">
+				<body>
+					<trans-unit id="9204248378636247318" datatype="html">
+						<source>Document <x id="PH" equiv-text="status.filename"/> was added to <x id="PH2" equiv-text="status.filename2"/> paperless.</source>
+ 					 	<context-group purpose="location">
+    					<context context-type="sourcefile">src/app/app.component.ts</context>
+    					<context context-type="linenumber">51</context>
+  						</context-group>
+  					<target state="needs-translation">Document <x id="PH" equiv-text="status.filename"/> was added to paperless.</target>
```
kaaposc, line 141: don't you need "PH2" also here?

pkg/convert/xliff12_test.go
DarjaLeonova: image
This is an example of a test case from MF2 Builder. So It will work with a dot.
This test case was used from the PDF file that @VladislavsPerkanuks provided.

VladislavsPerkanuks: we Could also make use of openning and closing functions. They are used for markup.

e.g.

```go
Message: "Document {+x :id=PH equiv-text=status.filename}{-x} was added to paperless"
```

but dunno about that 🤷

janishorsts: What is the purpose of ` id=\"PH\"`? Does id contain a value from a canonical list?

VladislavsPerkanuks: who knows :)

janishorsts: Do we know where the example is coming from?

janishorsts:
Too many unknowns.

PR #239, opened 2024-01-23
229-rename-all-pot-references-to-po

DarjaLeonova:


VladislavsPerkanuks (approved):

pkg/convert/po.go
```diff
@@ -107,10 +107,10 @@ func isOriginalPO(po pot.Po, override *bool) bool {
 	}
 
 	// NOTE: originality can also be determined by file extension.
-	// .pot == original, .po == translation.
+	// .po == original, .portableObject == translation.
```
klavinsjanis, line 110: .pot == original and .po == translation ?

cmd/client/client_integration_test.go
```diff
@@ -348,7 +348,7 @@ func Test_TranslationFileUpload_CLI(t *testing.T) {
 		})
 
 		require.ErrorContains(t, err,
-			"must be one of: json_ng_localize, json_ngx_translate, go, arb, pot, xliff_12, xliff_2")
```
klavinsjanis, line 351: shouldn't we allow .pot files as well?

klavinsjanis: nevermind, we are just supplying schema type.

PR #242, opened 2024-01-24
241-improve-go-caching-in-earthly

DarjaLeonova:


Earthfile
```diff
@@ -164,7 +167,7 @@ lint:
 
 test-unit:
   FROM +go
-  RUN --mount=type=cache,target=/go/pkg/mod go test ./...
+  RUN --mount=type=cache,id=mod-cache,target=/go/pkg/mod go test ./...
```
janishorsts, line 170: Please rename id to `go-mod` instead of `mod-cache`. Also, we can add go build.

```suggestion
  RUN \
    --mount=type=cache,id=go-mod,target=/go/pkg/mod \
    --mount=type=cache,id=go-build,target=/root/.cache/go-build \
    go test ./...
```

VladislavsPerkanuks (approved):
did it become faster ? :)

janishorsts (approved):

PR #243, opened 2024-01-24
224-po-lexer-quotes-must-be-present

klavinsjanis:


pkg/po/lex.go
```diff
@@ -187,26 +191,49 @@ func parseToken(line string, tokenType TokenType) (*Token, error) {
...
+			return strings.TrimPrefix(strings.TrimSuffix(fields[1], "\""), "\""), nil
+		}
+
+		return "", fmt.Errorf("invalid syntax in line string for token type: '%d'", tokenType)
+	case TokenTypeHeaderLanguage, TokenTypeHeaderTranslator, TokenTypeHeaderPluralForms,
+		TokenTypeHeaderProjectIDVersion, TokenTypeHeaderPOTCreationDate, TokenTypeHeaderPORevisionDate,
+		TokenTypeHeaderLanguageTeam, TokenTypeHeaderLastTranslator, TokenTypeHeaderXGenerator,
+		TokenTypeHeaderReportMsgidBugsTo, TokenTypeHeaderMIMEVersion, TokenTypeHeaderContentType,
+		TokenTypeHeaderContentTransferEncoding, TokenTypeHeaderGeneratedBy:
+		if len(fields) == n {
```
janishorsts, line 220: What is the case when `len(fields) != n`? Why do you ignore it?

janishorsts: From the following cases which ones are valid syntax?

```po
"Language:en-US\n"
" Language: en-US\n"
"Language: en-US \n"
```

klavinsjanis: if len(fields) != n empty string is returned.

janishorsts: Does `po` allow 1 or more whitespaces? Which cases have valid syntax?

```po
msgid "some text" <-- return `some text`
msgid  "some text" <-- returns empty string in your case
msgid "some text"
msgid"some text"
 msgid "some text"
```

janishorsts: ChatGPT:

msgid "some text": This is valid syntax. The keyword msgid is followed by a space and then the quoted string, which is the standard format.

msgid "some text": This is also valid syntax, although it's not standard. There are two spaces between msgid and the quoted string, but most parsers will handle this correctly. However, for consistency and readability, it's generally recommended to use only one space.

msgid"some text": This is not valid syntax. There should be a space between msgid and the quoted string. Without this space, the line may not be correctly parsed.

msgid "some text": Technically, this is valid syntax. The parser usually ignores leading whitespace, including the space before msgid. However, it's not standard or recommended practice to include this leading space, as it can affect readability and consistency in the file.

In summary, the first two lines are valid, with the second being non-standard due to the extra space. The third line is invalid due to the lack of space between msgid and the quote. The fourth line, while technically valid, is non-standard due to the leading space. For best practices, it's recommended to follow the format shown in the first line.

klavinsjanis: okay will add some tests

janishorsts: @klavinsjanis what does po spec says about it?

klavinsjanis: Is there formal specification?

pkg/po/lex_test.go
```diff
@@ -219,26 +219,98 @@ displayed in the filter.`),
...
+			name:     "msgctxt comment, value enclosed in spaces",
+			input:    `#| msgctxt     context    `,
+			expected: []Token{mkToken(TokenTypeMsgctxtPreviousContext, "msgctxt context")},
+		},
 		// negative tests
 		{
-			name: "When msgid value is incorrect",
+			name:        "language header, missing closing quotes",
+			input:       `"Language: en-US`,
+			expectedErr: fmt.Errorf("line must end with double quotation mark - \" "),
```
VladislavsPerkanuks, line 272: Calling formatting function without formatting itself.
I Suggest using errors.New() then

pkg/po/lex_test.go
VladislavsPerkanuks, line 1: do we really need so many negative tests ? 🤔 
what happened on focusing on happy path

pkg/po/lex.go
```diff
@@ -187,26 +216,78 @@ func parseToken(line string, tokenType TokenType) (*Token, error) {
...
+			value = " " + value
+		}
+
+		switch tt { //nolint:exhaustive
+		case TokenTypeMsgidPluralPrevUntStrPlural:
+			value = "msgid_plural" + value
+		case TokenTypeMsgctxtPreviousContext:
+			value = "msgctxt" + value
+		case TokenTypeMsgidPrevUntStr:
+			value = "msgid" + value
```
VladislavsPerkanuks, line 284: why this is here ? 🤔 
parseTokenValue extracts the value of the token, but here the value is also prefixed with tokenType.
1. I don't think it is correct
2. It contradicts with function godoc, particular with: `msgid "some text" -> some text`

klavinsjanis: well in the happy path tests -> 		mkToken(TokenTypeMsgidPrevUntStr, "msgid previous id"),

VladislavsPerkanuks: dont think that is correct :)

pkg/po/lex_test.go
klavinsjanis: is that many?

pkg/po/lex.go
```diff
@@ -90,84 +128,102 @@ func lex(r io.Reader) ([]Token, error) {
...
+		tokenType, linePrefix = TokenTypeHeaderGeneratedBy, headerGeneratedByPrefix
+	case strings.HasPrefix(line, msgCtxtPrefix):
+		tokenType, linePrefix = TokenTypeMsgCtxt, msgCtxtPrefix
+	case strings.HasPrefix(line, pluralMsgIDPrefix):
+		tokenType, linePrefix = TokenTypePluralMsgID, pluralMsgIDPrefix
+	case strings.HasPrefix(line, msgIDPrefix):
+		tokenType, linePrefix = TokenTypeMsgID, msgIDPrefix
+	case strings.HasPrefix(line, pluralMsgStrPrefix):
+		if tokenType, linePrefix = TokenTypePluralMsgStr,
+			regexp.MustCompile(`^msgstr\[(?:\d+|\*)\]`).FindString(line); linePrefix == "" {
```
VladislavsPerkanuks, line 176: I believe that compiling regex whenever the case is executed is not efficient. 
I would suggest moving this to package level variable

VladislavsPerkanuks (approved):
seems fine to me

DarjaLeonova (approved):

PR #248, opened 2024-01-29
update dependencies

DarjaLeonova:


mvilks (approved):

pkg/convert/po.go
```diff
@@ -245,9 +245,9 @@ func ToPo(t model.Translation) ([]byte, error) {
 			switch p := p.(type) {
 			case ast.TextPattern:
 				text += string(p)
-			case ast.PlaceholderPattern:
+			case ast.Pattern:
 				//nolint:forcetypeassert // No other types are possible.
-				text += placeholders[p.Expression.(ast.VariableExpression).Variable]
+				text += placeholders[p.(ast.VariableExpression).Variable]
```
VladislavsPerkanuks, line 250: What's the point of asserting interface type to the interface? I am curious why golang allows doing that :)

```go
case ast.VariableExpression: // No other types are possible for now
  text += placeholders[p.Variable]
```

pkg/fuzzy/google_translate.go
```diff
@@ -271,7 +271,7 @@ func patternToString(pattern []ast.Pattern) string {
 		switch v := pattern[i].(type) {
 		case ast.TextPattern:
 			text += string(v)
-		case ast.PlaceholderPattern:
+		case ast.Pattern:
 			text += fmt.Sprintf("{$%d}", i)
```
VladislavsPerkanuks, line 275: Same here
```suggestion
		case ast.LiteralExpression, ast.AnnotationExpression, ast.VariableExpression, ast.Markup: // or default
			text += fmt.Sprintf("{$%d}", i)
		}
```

VladislavsPerkanuks: Update godoc, there is no more `PlaceholderPattern`.
Patterns are:
- Text
- All Expressions (Literal, Variable, Annotation)
- Markup

go.sum
VladislavsPerkanuks, line 1: Something is off here.
Not the first time I seen this, in your PR's

When you work there are many additions to `go.sum` file, but when I run `go mod tidy` all those lines get deleted.
So this file always gets changed back and forth. 

So would be nice, if some 3rd party could run `go mod tidy`, and see which behavior is correct. And investigate why this inconsistency is appearing, would be nice too

VladislavsPerkanuks (approved):

klavinsjanis (approved):

pkg/fuzzy/google_translate.go
```diff
@@ -271,7 +271,7 @@ func patternToString(pattern []ast.Pattern) string {
 		switch v := pattern[i].(type) {
 		case ast.TextPattern:
 			text += string(v)
-		case ast.PlaceholderPattern:
+		case ast.LiteralExpression, ast.AnnotationExpression, ast.VariableExpression, ast.Markup: // or default
```
VladislavsPerkanuks, line 274: that default comment was for you :), shouldnt be here
either
1. case ast.LiteralExpression, ast.Annotation...
2. default

VladislavsPerkanuks (approved):

PR #255, opened 2024-02-07
254-MF2-to-ARB

DarjaLeonova:


pkg/convert/arb.go
```diff
@@ -141,11 +142,32 @@ func ToArb(translation model.Translation) ([]byte, error) {
 	// dst length = number of messages + number of potential descriptions (same as number of messages) + locale.
 	dst := make(map[string]interface{}, len(translation.Messages)*2+1)
 
+	patternsToMsg := func(patterns []ast.Pattern) string {
```
VladislavsPerkanuks, line 145: i think this will be also used in other converts, so might be good idea to move it outside this function

pkg/convert/arb.go
```diff
@@ -141,11 +142,32 @@ func ToArb(translation model.Translation) ([]byte, error) {
...
+		tree, err := ast.Parse(msg.Message)
+		if err != nil {
+			return nil, fmt.Errorf("parse mf2 message: %w", err)
+		}
+
+		if simpleMessage, ok := tree.Message.(ast.SimpleMessage); ok {
+			msg.Message = patternsToMsg(simpleMessage.Patterns)
+		}
+
+		dst[msg.ID] = msg.Message
```
VladislavsPerkanuks, line 170:
```suggestion
		if simpleMessage, ok := tree.Message.(ast.SimpleMessage); ok {
			dst[msg.ID] = patternsToMsg(simpleMessage.Patterns)
		}

returns errors.New(...) // complex is not implemented yet
```

VladislavsPerkanuks (approved):

pkg/convert/arb.go
VladislavsPerkanuks: This would be better
```suggestion
		switch mf2Msg := tree.Message.(type) {
		case nil:
			dst[msg.ID] = ""
		case ast.SimpleMessage:
			dst[msg.ID] = patternsToMsg(mf2Msg.Patterns)
		case ast.ComplexMessage:
			return nil, fmt.Errorf("complex message not supported")
		}
```

kaaposc (approved):

PR #296, opened 2024-06-19
Condition the integration test step

kaaposc:


.github/workflows/ci.yml
```diff
@@ -35,6 +35,7 @@ jobs:
       - name: Test Unit
         run: earthly --ci +test-unit
       - name: Test Integration
+        if: ${{ github.actor != 'dependabot[bot]' }}
```
janishorsts, line 38: Does it mean this step is skipped if the actor is 'dependabot[bot]' ?

